PR #5, opened 2015-11-19
Introduce grading

arbrandes:
For each <hastexo /> invocation in a lab vertical, the course author defines
one or more <test /> children that will check the state of the student
environment.  Each <test /> item contains code to be run by the public facing
node of the student environment.  The script can be written in any language, as
long as there is an installed interpreter for it.  The script must return 0 if
the check is successful, 1 if it fails.  For example:

```
<hastexo ...>
  <test>
    #!/bin/bash
    ...
  </test>
  <test>
    #!/usr/bin/env python
    ...
  </test>
</hastexo>
```

When the student clicks on "Check progress", the XBlock will copy each
assessment's code to a temporary file on the public facing node, make it
executable, and run it.

Student scores are published when all tests have been run.

For the moment, it is not possible to edit the tests in Studio.  They must be
defined as XML.

### DIFF
--- a/hastexo/hastexo.py
+++ b/hastexo/hastexo.py
@@ -1,11 +1,13 @@
 import json
 import logging
+import textwrap
 
 from .tasks import launch_or_resume_user_stack as launch_or_resume_user_stack_task
 from .tasks import suspend_user_stack as suspend_user_stack_task
+from .tasks import check as check_task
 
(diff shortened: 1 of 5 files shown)


fghaas:
This looks really slick overall. Let's please get this deployed to our test system as soon as possible. Thanks!

PR #6, opened 2015-12-03
Add Keystone v3 support

arbrandes:
This adds optional Keystone v3 support.  It tries to detect which
version (v2 or v3) is supported by the cloud, and uses the appropriate
one.

It is fully backwards-compatible with previous XBlock definitions.


fghaas:
Whoa, awesome. :)

fghaas:
Very neat stuff. Thanks!

PR #7, opened 2015-12-11
Don't wait forever for tasks to start running

arbrandes:


fghaas:
This LGTM, though I admittedly don't really feel qualified to review this, so take it with a pinch of salt. :)

PR #9, opened 2015-12-23
Do away with infinite loops

arbrandes:
There were a couple of edge cases where tasks would loop indefinitely.
For example, if a stack was deleted while the task waited for it to be
created or resumed.

This changeset treats this case specifically, but also gets rid of any
infinite waiting by introducing timeouts where there weren't any before.


fghaas:
This sounds like it must have been pretty painful to find and fix. Thank you!

arbrandes:
Yeah, and this is just part of the issue.  I've also discovered task concurrency isn't working.  In other words, when two people request stacks, the second one will only start when the first one is complete.  I'll create a new issue to track it.

PR #12, opened 2016-02-01
Use correct scope to store stack state

arbrandes:
The student's stack is the same across uses of the hastexo XBlock in a
course.  As such, certain state variables should be using the
appropriate scope for this.

The closest thing to what's needed is the "preferences" scope.  However,
the documentation states that it stores information across courses as
well.  It remains to be seen if this is actually the case.

### DIFF
--- a/hastexo/hastexo.py
+++ b/hastexo/hastexo.py
@@ -105,27 +105,29 @@ class HastexoXBlock(StudioEditableXBlockMixin, XBlock):
         scope=Scope.settings,
         help="Defines the maximum total grade of the block.")
 
-    # Scope: user state.  These are set per instance, per user.
-    user_stack_name = String(
-        default="",
-        scope=Scope.user_state,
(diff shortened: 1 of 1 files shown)


arbrandes:
This seems to fix the issue observed earlier by myself and @armaan.

arbrandes:
Yeah, it's not ideal, but there is no predefined scope for "block type, in a course run, per user".  There is an unnamed scope for "block definition, in a course run, per user", but 1) it's not usable from the XBlock because the API doesn't predefine it, and 2) the actual meaning of "block definition", and how it can be taken advantage of, is not clear.

The only alternative is to whip up a model for the XBlock, and use it instead of the XBlock state engine.  It's not unheard of (problem-builder does it), but is probably overkill just for this.

Anyway, since it's working, I'm merging it in, and we can revisit it when the need arises.

PR #13, opened 2016-02-05
Delete failed stacks and add unit tests for tasks

arbrandes:


fghaas:
I'm seeing an issue here where a stack is in a `RESUME_FAILED` state, but the user gets zero feedback and sees the "please wait..." dialog indefinitely. Please revisit and see if you can fix that before merging.

arbrandes:
I've added a specific unit test for RESUME_FAILED in the backend, and
made sure it is handled by the javascript (where there are no tests
yet), and it looks like the logic is correct.

Given that I've seen this before with different stack states (even with
successful ones), and from previous cursory investigations, my theory is
that we're being plagued by hung Celery processes.  An easy way to
ascertain this is to try and stop the worker processes when the user
isn't getting proper feedback.  If it hangs...

This is not new to this patch, btw.  But whatever the case, I'm going to
try and reproduce the problem and try to get a stack trace to find out
what might be going on.

fghaas:
OK, thanks!

arbrandes:
This PR now includes markdown lab instructions, as per story #174.

PR #15, opened 2016-04-26
Handle task timeouts more robustly

arbrandes:
Introduce code to handle a configurable task timeout (default = 10
minutes) at the highest possible level: during each browser request.

This should help handle corner cases where the task ID is not cleared
when a task never returns, thus making the XBlock think that a
particular stack is being launched forever.

### DIFF
--- a/hastexo/hastexo.py
+++ b/hastexo/hastexo.py
@@ -2,9 +2,11 @@ import json
 import logging
 import textwrap
 import markdown2
+import datetime
+import pytz
 
 from xblock.core import XBlock
-from xblock.fields import Scope, Integer, Float, String, Dict, List
+from xblock.fields import Scope, Integer, Float, String, Dict, List, DateTime
 from xblock.fragment import Fragment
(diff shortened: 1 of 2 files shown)


hastexo/hastexo.py
```diff
@@ -273,6 +284,7 @@ def launch_or_resume_user_stack(self, sync = False):
         else:
             result = task.apply_async(args=args, kwargs=kwargs, expires=60)
             self.user_stack_launch_id = result.id
+            self.user_stack_launch_timestamp = datetime.datetime.now(pytz.utc)
```
fghaas: Minor note here (not for now, but to be revisited later): I find it a little weird that a timestamp would need to know its applicable timezone.

arbrandes: That's Python for ya:  http://stackoverflow.com/questions/796008/cant-subtract-offset-naive-and-offset-aware-datetimes

The XBlock DateTime uses the pytz.utc trick, so we have to, as well.

PR #16, opened 2016-04-29
Don't delete stacks indiscriminately

arbrandes:
Assume a rollback stance when stack launching fails: if the failure
occurred during creation, delete it; during resuming, suspend it.

For clarity, also separate the stack launch status from the verification
status as much as possible.

TG-3 #closed

### DIFF
--- a/hastexo/tasks.py
+++ b/hastexo/tasks.py
@@ -2,6 +2,7 @@ import time
 import os
 import paramiko
 import uuid
+import re
 
 from celery import Task
 from celery.utils.log import get_task_logger
@@ -29,6 +30,7 @@ class LaunchStackTask(Task):
         """
 
         status = None
+        verify_status = None
         error_msg = None
         stack = None
         stack_ip = None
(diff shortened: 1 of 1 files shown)


hastexo/tasks.py
```diff
@@ -99,7 +107,8 @@ def launch_stack(self, heat, stack_name, stack_template):
             status = stack.stack_status
             retry += 1
             if retry >= self.retries:
-                status = 'CREATE_FAILED'
+                status_prefix = status[:-12]
```
fghaas: Not sure where the `-12` comes from. Also, would probably be nicer (and more maintainable) to use a regex match/replace here.

fghaas:
It would be really helpful if the failure messages would include the stack UUID on top of the stack name. Once a stack is deleted, it's non-trivial to find out what its UUID was, but only the UUID would show up in the cloud provider's logs.

hastexo/tasks.py
arbrandes: -12 is to remove the suffix from *_IN_PROGRESS, the only possible statuses here.  But ok, a regex sub will do just as well.

arbrandes:
While the last comment makes perfect sense, this warrants a separate user story (or extending TG-1).  It's not always trivial to get the stack UUID from within the XBlock, either.

hastexo/tasks.py
arbrandes: Fixed on latest PR.

fghaas:
Sure, putting stack UUID logging into its own US is fine.

PR #23, opened 2016-10-08
v0.2.1

arbrandes:


arbrandes:
I've seen that before.  It's just a different flavor of the refused connection problem. :/

fghaas:
Ack. FTR it did work on the second try.

hastexo/hastexo.py
```diff
@@ -266,9 +267,12 @@ def launch_or_resume_user_stack(self, sync = False):
         kwargs = self._get_os_auth_kwargs()
         task = LaunchStackTask()
         if sync:
+            logger.info('Running synchronous launch task for [%s]' % (self.stack_name))
```
fghaas: Out of curiosity: whenever convenient, could you explain when stack launches happen synchronously, and when asynchronously?

arbrandes:
Bear with me as I give some background before actually answering the
question.

If edX ran behind an event-loop server such as Tornado, they'd all be
synchronous calls, as it would be faster from the point of view of the
user: there would be no additional wait time for the last poll.  As it
stands, though:

* It's too expensive, server side, to keep a connection open for long.
Because of this, nginx will kill connections after 60 seconds by default.
 (We've actually been hitting this on these connection issues, hence the
improved error handling on this PR.)
*  It wouldn't make too much of a difference, time-wise, if launches were
synchronous, because the server still needs to poll Heat every X seconds.
 (Ideally, there'd be a pub-sub Heat API, or the TCP connection would be
kept open until the stack changed state.)

Hence, most communication with Heat is handled asynchronously, by Celery
workers.  But there is an edge case that needs to be handled.  Consider the
following sequence of events, in chronological order:

1) User opens a hastexblock unit.
2) XBlock fires an async Celery task to launch a stack, and quits.
3) 10 seconds later, the user's browser asks how the launch is going
(meanwhile, the user sees the "Please wait" message).  The Xblock queries
memcache to see if there is a result from the original launch task.  There
is none, so it returns "Pending" to the browser.  The launch task is still
running, polling Heat every X seconds to see if the task is ready.
4) The above happens a few times, the user gets tired of waiting, and
closes the tab.
5) While the user is AFK, the launch finishes.  The successful result is
stored in Memcached.
6) Soon afterwards, the stack is suspended by the dead man's switch, since
the browser isn't sending keepalives.
7) The user comes back 20 minutes later and opens the hastexblock unit
again.  Memcached is queried synchronously, and tells the XBlock that the
launch was successful, as per step 5).

Here's the rub.  If I tell the browser the launch was successful, it's fire
up Gateone and try to connect... to a suspended stack.  No good.  So I need
to make absolutely sure, again and as fast as possible, that the stack is
up.  Why as fast as possible?  Because we don't want to penalize the user
that was patient and *didn't* close the tab.  So we run this last launch
synchronously, AND avoid scheduling a suspend task while it is happening.
That way, the user is sure to be able to connect once it completes.

"But don't the same restrictions regarding cost and nginx timeout still
apply?", you might ask.  Yes, of course.  But at this point, we can
reasonably expect that either the stack is up, and the last check will take
just a few seconds, or the stack is suspended, and the task will take a
little longer but still less than a minute.

As mentioned above, though, the nginx timeout may still be hit, as during
the sync launch task we may be throttled by City for more than 60 seconds.

That's about it.  I wrote this up because you might have better ideas as to
how to handle this.  I certainly intend to improve the algorithm as part of
our "Speed up time to terminal" story.

PR #24, opened 2016-10-10
v0.2.2

arbrandes:
This is a consequence of https://github.com/hastexo/hastexo-xblock/pull/23#issuecomment-252611838.  In investigating a way to do away with the synchronous request, I found a few related bugs and fixed them.


fghaas:
This all looks excellent! Could you just please swap the order of the last two commits, so that the commit that bumps the version to 0.2.2 is also the one tagged v0.2.2?

arbrandes:
They are swapped - but github apparently lists commits by date, regardless of actual git order.  Weird.

PR #25, opened 2016-10-19
v0.2.3

arbrandes:


hastexo/hastexo.py
```diff
@@ -341,10 +341,20 @@ def student_view(self, context=None):
...
+            frag = Fragment(html)
+            frag.add_css_url(self.runtime.local_resource_url(self, 'public/css/main.css'))
+            return frag
+
         # Load the stack template from the course's content store
-        loc = StaticContent.compute_location(course_id, self.stack_template_path)
-        asset = contentstore().find(loc)
-        self.stack_template = asset.data
+        try:
+            loc = StaticContent.compute_location(course_id, self.stack_template_path)
```
fghaas: Does this still qualify as "naughty"?

arbrandes:
Yes, very much so.  Not way around it, though.  There's no API for
fetching stuff from the content store.

fghaas:
All right. Apparently this, and `instructions_path` (now deprecated) are the only instances left where we are access the content store.

Now, if we instead allowed the XBlock to optionally retrieve the template from a URL, like the Heat client already does, and only fall back on the content store if the URL weren't absolute, wouldn't that make it less naughty?

As in:
```python
import urlparse     # urllib.parse for python3

template_location = urlparse.urlparse(self.stack_template_path)
if template_location.netloc:
    # absolute URL, fetch template, don't touch content store
    # [...]
else:
    # relative URL, get template from content store
    try:
        loc = StaticContent.compute_location(course_id, self.stack_template_path)
        # [...]
```
Would that make sense?

arbrandes:
If the intent is to avoid this particular naughtiness, we can simply
fetch the template via HTTP from the courseware, rather than from mongo
directly.  That would also allow for fetching templates from external
URLs.

(FYI, hijacking Celery is even naughtier, but there's no easy way around
that.)

PR #27, opened 2016-11-05
Fix status detection

arbrandes:


hastexo/hastexo.py
```diff
@@ -444,7 +444,8 @@ def _process_error(error_msg):
             status = _process_result(result)
 
         # The stack was previously launched successfully
-        elif "COMPLETE" in last_status_string:
+        elif (last_status_string == "CREATE_COMPLETE" OR
+              last_status_string == "RESUME_COMPLETE"):
```
fghaas: I think the Pythonic way of doing this is
```python
elif last_status_string in ["CREATE_COMPLETE", "RESUME_COMPLETE"]
```
... but I may be totally bikeshedding here.

arbrandes: Good call.  Your comment inspired me to Do This Properly (tm).  See latest push.

fghaas:
LGTM now, thanks!

PR #35, opened 2017-03-14
v0.5.0

arbrandes:


fghaas:
Slick! Has this been tested against both a Keystone v3 and a v2 environment?

arbrandes:
I tested it against City and Elastx, so, presumably, yes.

fghaas:
Great. Will you publish to PyPI, as usual?

arbrandes:
Done: https://pypi.python.org/pypi/hastexo-xblock/0.5.0

fghaas:
Thanks!

PR #49, opened 2017-06-23
v0.5.4

arbrandes:


fghaas:
Uh, that build failure is weird. I presume you ran `tox` locally?

setup.py
```diff
@@ -30,6 +30,38 @@ def package_data(pkg, roots):
...
+def is_requirement(line):
+    """
+    Return True if the requirement line is a package requirement;
+    that is, it is not blank, a comment, or editable.
+    """
+    # Remove whitespace at the start/end of the line
+    line = line.strip()
+
+    # Skip blank lines, comments, and editable installs
+    return not (
```
fghaas: Not that it's very important, but you probably want to do this [with a regexp match](https://docs.python.org/2/library/re.html) instead — no?

arbrandes: Loading setup.py requirements from a file was tripping up tox, so I ditched it for now, in the interest of being able to redeploy everything sooner rather than later.

arbrandes:
Merging this so I can deploy it.

PR #52, opened 2017-09-20
[WIP] v2.0.0

arbrandes:


fghaas:
Now that we named this version 2.0, perhaps we should fork off a `stable-0.5` branch and have `master` eventually become `stable-2.0`? (We'll skip version 1, which is OK by me, considering the pivot from GateOne to guac.)

fghaas:
I just noticed that the checks that fail (besides a handful of trivial flake8 complaints) are all related to Python 2.7. Do we even want/need to support Python 2 anymore?

fghaas:
I think this is good to merge as soon as the tests pass. Let me know what you think about my earlier suggestion about a `stable-0.5` branch.

fghaas:
OK so the Python 3 tests are still full of errors but as far as I can tell those are all in upstream xblock and xblock-sdk, so probably nothing we can do about those.

One more flake8 nit (which Travis flags, though I did double-check and I didn't see it in my own build environment even with `tox -r`, so I'm not entirely sure what's up there):

```
 ./hastexo/tasks.py:452:69: F841 local variable 'e' is assigned to but never used
```

PR #65, opened 2018-04-12
v2.2.0

arbrandes:


hastexo/utils.py
```diff
@@ -21,6 +24,8 @@
     "suspend_timeout": 120,
     "suspend_interval": 60,
     "suspend_concurrency": 4,
+    "delete_age": 14,
```
fghaas: I'm never a big fan of mixing time units, because that sort of thing tends to get very confusing. I suggest this should be defined in seconds as well (even though that means we'll get pretty big numbers).

hastexo/management/commands/undertaker.py
```diff
@@ -0,0 +1,26 @@
+from django.core.management.base import BaseCommand
+from django.conf import settings as django_settings
+from apscheduler.schedulers.blocking import BlockingScheduler
+
+from hastexo.utils import SETTINGS_KEY, DEFAULT_SETTINGS
+from hastexo.jobs import UndertakerJob
+
+
+class Command(BaseCommand):
+    help = 'Deletes stacks automatically'
```
fghaas: Perhaps change this to "Automates the deletion of stale stacks, i.e. those that have not been resumed in a configurable time period" to be a little more descriptive.

tests/unit/test_jobs.py
```diff
@@ -38,6 +39,11 @@ def setUp(self):
             "suspend_timeout": 120,
             "suspend_concurrency": 1,
             "suspend_in_parallel": False,
+            "delete_age": 14,
+            "task_timeouts": {
+                "sleep": 0,
+                "retries": 10
```
fghaas: Tiny nit: not sure why this uses a plural here, should probably be `retry`.

hastexo/jobs.py
```diff
@@ -91,10 +94,117 @@ def suspend_stack(self, stack):
...
+            # Delete stack
+            heat_client.stacks.delete(stack_id=stack.name)
+
+            # Save status
+            stack.status = DELETE_IN_PROGRESS_STATE
+            stack.save()
+
+            # Wait until delete finishes.
+            retry = 0
+            while ('FAILED' not in stack.status and
```
fghaas: So if I read this correctly, this would retry for as long as Heat reports the stack as `DELETE_IN_PROGRESS`, and succeed if Heat either reports `DELETE_COMPLETE`, or issues a 404. However, it would stop if Heat reported `DELETE_FAILED`. Shouldn't it retry then, as well? (Consider that stack deletion can take several minutes, during which a compute node might die, or the Heat API may suffer an interruption, or `heat-engine` could be restarted due to maintenance.)

PR #66, opened 2018-05-18
v2.3.0

arbrandes:


hastexo/hastexo.py
```diff
@@ -598,7 +604,7 @@ def _process_result(result):
                         'error_msg': 'Unexpected result: %s' % repr(result.result)  # noqa: E501
                     }
             else:
-                status = {'status': 'PENDING'}
+                status = {'status': 'CHECK_PENDING'}
```
fghaas: Semantic note: there's a "check" operation that the Heat API supports (see https://developer.openstack.org/api-ref/orchestration/v1/#check-stack-resources, also `openstack stack check`), are you sure it doesn't clash with this? We might, at a later date, actually use the "real" check operation from the XBlock, even if we don't do so now.

To avoid the ambiguity from the get-go, maybe call this `PROGRESS_CHECK_PENDING`?

fghaas:
Looks great overall to me. Just one nit about naming/semantics.

PR #67, opened 2018-05-25
v2.3.1

arbrandes:


fghaas:
I am obviously absolutely no authority on Hawthorn, but this does looks good to me.

Maybe bump the minor, though, to reflect that this is built for a new upstream edX release, i.e. call it 2.4.0?

PR #69, opened 2018-06-07
v2.3.1

arbrandes:


arbrandes:
I've checked this works.  Upgraded tomcat8, rebooted, and fired up a lab.

PR #70, opened 2018-06-14
v2.3.2

arbrandes:


hastexo/tasks.py
```diff
@@ -364,6 +365,20 @@ def check_stack(self, stack, was_resumed):
                     paramiko.ssh_exception.SSHException,
                     paramiko.ssh_exception.NoValidConnectionsError):
                 self.sleep()
+            except Exception as e:
```
fghaas: Either do something with `e`, or make it just `except:` (PEP-8).

hastexo/tasks.py
```diff
@@ -364,6 +365,20 @@ def check_stack(self, stack, was_resumed):
...
+                if was_resumed:
+                    error_status = 'RESUME_FAILED'
+                    cleanup = 1
```
fghaas: Any chance we could get descriptive constant names here, instead of the rather opaque `1` and `2`?

hastexo/tasks.py
arbrandes: We don't want a bare `except:` because we really only want subclasses of `Exception`.  So I'll just drop the `as e`.

hastexo/tasks.py
arbrandes: Fair enough.

hastexo/tasks.py
fghaas: Oh, OK. Makes sense.

fghaas:
Good to merge for me. Thanks!

PR #79, opened 2018-10-18
v2.5.0

arbrandes:
Introduces multi-provider support.

### DIFF
--- a/README.md
+++ b/README.md
@@ -368,9 +368,6 @@ configured with the following attributes:
 * `protocol`: One of 'ssh', 'rdp', or 'vnc'.  This defines the protocol that
   will be used to connect to the environment.  The default is 'ssh'.
 
-* `provider`: (Optional) The name of an OpenStack provider configured in the
-  platform.
-
 * `stack_ports`: (Optional) A list of port numbers the user can choose from.
   This is intended as a means of providing a way to connect directly to
   multiple VMs in a lab environment, via port forwarding or proxying at the VM
@@ -380,6 +377,22 @@ configured with the following attributes:
   shown as a dropdown to the user.  It must match the list of port numbers in
   both order and number.
 
+You can also use the following nested XML elements:
+
+* `provider`: Reference to an OpenStack provider configured in the platform.
+  Its `name` attribute must match one of the providers in the XBlock
+  configuration, `capacity` specifies how many environments should be launched
+  in that provider at maximum (where "-1" means keep launching environments
+  until encountering a launch failure, and "0" disables the provider), and
+  `environment` optionally specifies a content store path to a Heat environment
+  file specifying template parameters for that provider.  This element must be
+  specified at least once, and can be specified multiple times to use one or
+  more fallback providers.
+
+* `test`: The contents of this element will be run verbatim as a script in the
+  user's lab environment, when they click the "Check Progress" button.  As
+  such, it should define an interpreter via the "shebang" convention.  This
+  element is optional, but can be specified multiple times.
 
 For example, in XML:
 
@@ -390,9 +403,19 @@ For example, in XML:
     stack_template_path="hot_lab.yaml"
     stack_user_name="training"
     protocol="rdp"
-    provider="default"
     stack_ports="[3389, 3390]"
-    stack_port_names="['server1', 'server2']" />
+    stack_port_names="['server1', 'server2']">
+    <provider name="provider1" capacity="20" environment="hot_env1.yaml" />
+    <provider name="provider2" capacity="30" environment="hot_env2.yaml" />
+    <provider name="provider3" capacity="0" environment="hot_env3.yaml" />
+    <test>
+      #!/bin/bash
+      # Check for login on vm1
+      logins=$(ssh vm1 last root | grep root | wc -l)
+      [ $logins -ge 1 ] || exit 1
+      exit 0
+    </test>
+  </hastexo>
 </vertical>
 ```
 
--- a/hastexo/hastexo.py
+++ b/hastexo/hastexo.py
@@ -9,16 +9,23 @@ from xblockutils.resources import ResourceLoader
 from xblockutils.studio_editable import StudioEditableXBlockMixin
 from xblockutils.settings import XBlockWithSettingsMixin
 
-from django.db import transaction
 from django.utils import timezone
 
-from .models import Stack
 from .utils import (UP_STATES, LAUNCH_STATE, LAUNCH_ERROR_STATE, SETTINGS_KEY,
-                    DEFAULT_SETTINGS, get_xblock_configuration)
+                    get_xblock_settings, update_stack, get_stack)
 from .tasks import LaunchStackTask, CheckStudentProgressTask
 
 logger = logging.getLogger(__name__)
 loader = ResourceLoader(__name__)
+
+
+class LaunchError(Exception):
+    error_msg = ""
+
+    def __init__(self, error_msg):
+        super(LaunchError, self).__init__()
+
+        self.error_msg = error_msg
 
 
 @XBlock.wants('settings')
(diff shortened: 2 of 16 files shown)


fghaas:
Ah no worries, I won't. :)

arbrandes:
Ok, this is ready for testing and code review.

README.md
```diff
@@ -380,6 +377,17 @@ configured with the following attributes:
   shown as a dropdown to the user.  It must match the list of port numbers in
   both order and number.
 
+You can also use the following nested XML tags:
+
+* `provider`: (Optional) A list of OpenStack providers configured in the
```
fghaas: Minor nit: this sounds like the xblock expects a _single_ nested `<provider>` tag, which is then interpreted as a list. I'd suggest to rephrase this as:

```markdown
* `provider`: Reference to an OpenStack provider configured in the platform.
  It `name` attribute must match one of the providers in the XBlock configuration,
  `capacity` specifies how many environments should be launched in that
  provider at maximum (where 0 means keep launching environments until 
  encountering a launch failure), and 
  `environment` specifies a content store path to a Heat environment file specifying
  template parameters for that provider.
  This element must be specified at least once, and can be specified multiple times
  to use one or more fallback providers.
```

The part where it says "where 0 means launch until encountering a launch failure" is of course subject to the rest of the discussion, i.e. what does setting 0 mean, as opposed to not defining the attribute at all. Is it equivalent (meaning 0 means launch until failure), or is it 0 actually 0 (meaning launch nothing here at all, something that one might use to temporarily disable a provider in a deployed course) .

README.md
```diff
@@ -380,6 +377,17 @@ configured with the following attributes:
   shown as a dropdown to the user.  It must match the list of port numbers in
   both order and number.
 
+You can also use the following nested XML tags:
```
fghaas: Suggest to use "elements" instead of "tags" here.

README.md
```diff
@@ -380,6 +377,17 @@ configured with the following attributes:
...
+
+* `provider`: (Optional) A list of OpenStack providers configured in the
+  platform.  `name` must match one of the providers in the XBlock
+  configuration, `capacity` specifies how many simultaneous environments should
+  be launched in that provider before trying the next one, and `environment`
+  specifies a content store path to a Heat environment file specifying template
+  parameters for that provider.
+
+* `test`: (Optional) A list of scripts that will run against the lab
```
fghaas: Same as above. Suggest to clarify that you can define _multiple_ `<test>` elements which then _constitute_ a list of tests, rather than a _single_ `<test>` element that _contains_ a list of tests.

hastexo/hastexo.py
```diff
@@ -126,6 +133,17 @@ def parse_xml(cls, node, runtime, keys, id_generator):
                 text = textwrap.dedent(text)
 
                 block.tests.append(text)
+            elif child.tag == "provider":
+                name = child.attrib["name"]
+                if not name:
+                    raise KeyError("name")
+                capacity = child.attrib.get("capacity", None)
```
fghaas: Since this defaults to `None`, what is the defined behavior if the user sets `0`, or a negative number?

hastexo/hastexo.py
```diff
@@ -163,31 +181,25 @@ def get_block_ids(self):
...
             from xmodule.exceptions import NotFoundError
 
-            loc = StaticContent.compute_location(course_id,
-                                                 self.stack_template_path)
+            loc = StaticContent.compute_location(course_id, path)
             asset = contentstore().find(loc)
-            stack_template = asset.data
+            contents = asset.data
         except (ImportError, NotFoundError):
             pass
```
fghaas: Do we really want to succeed silently on failure here? Not even a log message if the course author fails to include a template, or botches the reference?

hastexo/hastexo.py
```diff
@@ -318,24 +299,63 @@ def launch_stack_task_result(self, task_id):
...
         def _launch_stack(reset=False):
-            args = (
-                configuration,
-                self.stack_run,
-                self.stack_name,
-                self.get_stack_template(),
-                self.stack_user_name,
-                reset
-            )
+            stack_template = self.read_from_contentstore(
```
fghaas: What happens if this read fails? I see no exception handler here, which is generally fine — we do want to bail if the template can't be found. But would we get a stack trace somewhere?

hastexo/hastexo.py
```diff
@@ -318,24 +299,63 @@ def launch_stack_task_result(self, task_id):
...
+                self.stack_template_path)
+            providers = []
+            if len(self.providers):
+                for provider in self.providers:
+                    p = dict(provider)
+                    env_path = p.get("environment")
+                    p["environment"] = self.read_from_contentstore(env_path)
```
fghaas: Same as above: how do we handle a read failure here?

hastexo/tasks.py
```diff
@@ -76,174 +74,332 @@ def run(self,
...
+
+                stack_data = self.try_all_providers()
+            else:
+                stack_data = self.try_provider(stack.provider)
+        else:
+            stack_data = self.try_all_providers()
+
+        return stack_data
+
+    def get_provider_stack_count(self, provider):
```
fghaas: I very much like the idea that instead of keeping track of a "stacks per provider" semaphore count, we just do a `SELECT COUNT` every time we need to get the stack count. We'll need to see how this scales, I suppose, but my hunch is that this scales far better than any contentious updates.

README.md
arbrandes: +1 for improving the wording.  As for 0 versus `None`, they're currently equivalent, in that they translate to infinite capacity.  However, you have a point:  we may want to use -1 for infinite, and 0 for disabled.  What do you think?

hastexo/tasks.py
arbrandes: Yup, my feelings exactly.  However, I don't _think_ this will bog down our mariadb cluster any time soon.  The `hastexo_stack` table has not yet gone beyond 10k records in any of our production stacks, and unless we're talking tens of millions of stacks, we can probably handle our worst cases with some further InnoDB tuning.

arbrandes:
I've addressed the comments in the latest fixup [commit](https://github.com/hastexo/hastexo-xblock/pull/79/commits/66eb7bad2c594ce064871bce4bbeb936d66f9893), which I'll squash later.

README.md
```diff
@@ -377,17 +377,22 @@ configured with the following attributes:
...
+  Its `name` attribute must match one of the providers in the XBlock
+  configuration, `capacity` specifies how many environments should be launched
+  in that provider at maximum (where "-1" means keep launching environments
+  until encountering a launch failure, and "0" disables the provider), and
+  `environment` optionally specifies a content store path to a Heat environment
+  file specifying template parameters for that provider.  This element must be
+  specified at least once, and can be specified multiple times to use one or
+  more fallback providers.
+
+* `test`: The contents of this element will be run verbatum as a script in the
```
fghaas: It's "verbatim". :)

README.md
```diff
@@ -377,17 +377,22 @@ configured with the following attributes:
...
+  user's lab environment, when the they click the "Check Progress" button.  As
+  such, it should define an interpreter via the "shabang" convention.  This
```
fghaas: Commonly "shebang".

hastexo/tasks.py
```diff
@@ -175,12 +175,22 @@ def try_all_providers(self):
...
-
-            if capacity:
+            capacity = provider.get("capacity")
+            if capacity in (None, "None"):
+                capacity = -1
+            else:
+                try:
+                    capacity = int(capacity)
+                except (TypeError, ValueError):
+                    logger.info("Provider [%s] has invalid capacity." %
```
fghaas: `.info()` seems weak here. This should be a warning or an error, I think.

arbrandes:
All good points. :)  Addressed [here](https://github.com/hastexo/hastexo-xblock/pull/79/commits/584d6663e26c3981f4b6f191aec21a5a2a1ae0b0)

arbrandes:
Any objections to merging this in?

fghaas:
None, go ahead!

PR #83, opened 2018-11-20
v2.5.3

arbrandes:


tests/unit/test_tasks.py
```diff
@@ -183,6 +183,195 @@ def test_create_stack(self):
...
+        self.assertEqual(res["status"], "CREATE_COMPLETE")
+        self.assertEqual(res["provider"], self.providers[1]["name"])
+        self.assertEqual(self.get_stack("provider"), self.providers[1]["name"])
+        self.assertEqual(res["error_msg"], "")
+
+    def test_heat_error_on_all_providers(self):
+        # Setup
+        heat = self.get_heat_client_mock()
+        heat.stacks.get.side_effect = [
+            HTTPException,
```
fghaas: Instead of testing with generic `HTTPException` each time, perhaps cycle through some of its various subclasses (such as `HTTPBadRequest`, `HTTPForbidden`, `HTTPServiceUnavailable`) so we make sure we handle all of them properly?

Applies to other tests too.

README.md
```diff
@@ -402,12 +399,12 @@ For example, in XML:
...
     stack_template_path="hot_lab.yaml"
     stack_user_name="training"
-    protocol="rdp"
-    stack_ports="[3389, 3390]"
-    stack_port_names="['server1', 'server2']">
+    protocol="rdp">
     <provider name="provider1" capacity="20" environment="hot_env1.yaml" />
     <provider name="provider2" capacity="30" environment="hot_env2.yaml" />
     <provider name="provider3" capacity="0" environment="hot_env3.yaml" />
+    <port name="server1" number="3389" />
```
fghaas: Completely on board with this change; much nicer this way.

README.md
```diff
@@ -368,50 +368,68 @@ configured with the following attributes:
...
-    <port name="server1" number="3389" />
-    <port name="server2" number="3390" />
-    <test>
-      #!/bin/bash
-      # Check for login on vm1
-      logins=$(ssh vm1 last root | grep root | wc -l)
-      [ $logins -ge 1 ] || exit 1
-      exit 0
-    </test>
+    <option:providers>
```
fghaas: OK so YAML in XML looks rather like eeek to me, _particularly_ if for tests we now have to wrap bash in YAML wrapped in XML (jush bash in XML was already bad enough, with CDATA sections required for `&`, for example) plus this will be an awful lot of rewrites in existing courses, so don't expect to be able to rip the deprecated bits out anytime soon.

If you're convinced that this is a good idea moving forward (considering you're buying long-term improved maintainability with medium-term parallel maintenance), then please enumerate the advantages that you're seeing here.

tests/unit/test_tasks.py
arbrandes: All the other exceptions of this class are subclasses of HTTPException, but I agree that it would make for more realistic tests.  On it.

README.md
arbrandes: I'm with you that for tests this is particularly ugly - though it's not as if it is currently pretty, either.  My thinking boils down to the following:

1. Maintainability: once (and if, of course) we drop the specialized parsing, it's that much less code to maintain and test.
2. Simplicity, consistency, cheaper development: it'll be easier to add fields like these to the XBlock in the future, with no need to parse them (and unit test the parsing) in a special way.
3. Exports: exported OLX of option fields is less awful than what we have currently, in that the JSON list is rendered as tag content with no escaping (versus an ugly-ass escaped string in an attribute).  This will prove useful in maintaining courseware that is generated via Studio (versus olx-utils).
4. YAML: While YAML in XML is not pretty (and shell in YAML in XML is goldbergian), in practice we can use olx-utils to keep the YAML nicely separated from the XML.  The latter being the case, from the point of the course author it'll be YAML vs XML, and we know which one is better for humans.  (I.e., `<test>#/bin/bash</test>` is worse than `- | #/bin/bash`.)

However, what I may not have made clear is that one thing doesn't exclude the other.  As things stand with this PR, the course author can decide whether to use XML or YAML-in-XML, or a mix of both.  Plus, while dropping custom XML parsing will improve maintainability, adding support for the option field has near-zero impact: it's a feature of the XBlock SDK, so we get it for free.  All we have to do is keep field names consistent between both paradigms.

In other words, would you be happier if I just dropped the deprecation language?  I admit _I'm_ not even convinced we should.

fghaas: > I'm with you that for tests this is particularly ugly - though it's not as if it is currently pretty, either.

No objection here at all.

> YAML: While YAML in XML is not pretty (and shell in YAML in XML is goldbergian), in practice we can use olx-utils to keep the YAML nicely separated from the XML

So basically you're suggesting that we'll be using *just* YAML, Jekyll-style, for lab definitions, and then use olx-utils to mangle that into XML with YAML? In other words, we'll have Markdown for lab instructions and YAML for the lab definition?

If that's what you're suggesting, then I'd count that as a real UX improvement that I'd be strongly in favor of — anything that lets course authors write _less_ XML is good.

> In other words, would you be happier if I just dropped the deprecation language?

No I think if there are two ways of doing something there should always be a preferred way and a less-preferred way (this is Python not Perl, after all), so I'm perfectly OK in principle with one of the two approaches being deprecated.

arbrandes: > we'll have Markdown for lab instructions and YAML for the lab definition?

With the current dumb olx-utils templates, we'd immediately support going from this:

```
<hastexo
  stack_template_path="hot_lab.yaml"
  stack_user_name="training"
  stack_protocol="ssh">
  <provider name="provider1" capacity="20" environment="hot_env1.yaml" />
  <provider name="provider1" capacity="30" environment="hot_env2.yaml" />
  <%block name="hastexo_tests" />
</hastexo>
<%block name="hastexo_tests"><%text>
  <test>
    #!/bin/bash
    exit 0
  </test>
  <test>
    #!/bin/bash
    exit 0
  </test>
</%text></%block>
```

To this:

```
<hastexo xmlns:option="http://code.edx.org/xblock/option"
  stack_template_path="hot_lab.yaml"
  stack_user_name="training"
  stack_protocol="ssh">
  <option:providers>
    - name: provider1
      capacity: 20
      environment: hot_env1.yaml
    - name: provider2
      capacity: 30
      environment: hot_env2.yaml
  </option:providers>
  <option:tests><%block name="hastexo_tests">[]</%block></option:tests>
</hastexo>
<%block name="hastexo_tests"><%text>
  - |
    #!/bin/bash
    exit 0
  - |
    #!/bin/bash
    exit 0
</%text></%block>
```

However, I believe it wouldn't be all that hard to modify the main lab template (`include/sequential/lab.xml`) to use PyYAML to parse external pure-YAML files and spit out the above.  (It would probably be even easier to spit out JSON, actually.  But at this point, it wouldn't matter to the author, and these fields take JSON just as well as they do YAML.)

In the latter case, I _think_ we could manage something like:

```
include/sequential/lab.yaml

hastexo:
  stack_template_path: hot_lab.yaml
  stack_user_name: training
  stack_protocol: ssh
  providers:
    - name: provider1
      capacity: 20
      environment: hot_env1.yaml
    - name: provider2
      capacity: 30
      environment: hot_env2.yaml

sequential/lab_x.yaml

hastexo:
  tests:
    - |
      #!/bin/bash
      exit 0
    - |
      #!/bin/bash
      exit 0
```

tests/unit/test_tasks.py
arbrandes: See [this fixup](https://github.com/hastexo/hastexo-xblock/pull/83/commits/0b68b4ff1b3ed5f0eeeeebe74af0f97798bd4f16)

README.md
fghaas: ... and also generate reveal.js content from YAML/md? I'd not dislike that at all. :)

arbrandes: Yup, the whole shebang. :)  I'll give it a think and sketch up a proposal.

arbrandes: See the proposal here: https://github.com/hastexo/olx-utils/issues/22

fghaas:
OK I think we're good here. Please squash the fixup commits and force-push, and then this is good to merge. Thanks!

PR #84, opened 2018-12-11
v2.5.4

arbrandes:


hastexo/jobs.py
```diff
@@ -151,6 +153,8 @@ def run(self):
                 suspend_timestamp__lt=cutoff
             ).exclude(
                 status__in=dont_delete
+            ).exclude(
+                provider__exact=''
```
fghaas: Under what circumstances would this happen? In other words, when is the provider ever empty?

hastexo/jobs.py
```diff
@@ -90,37 +90,43 @@ def suspend_stack(self, stack):
...
+                heat_stack = heat_client.stacks.get(stack_id=stack.name)
+            except HTTPNotFound:
+                heat_status = DELETED_STATE
+            else:
+                heat_status = heat_stack.stack_status
 
-            # Suspend stack
-            heat_client.actions.suspend(stack_id=stack.name)
+            if heat_status in UP_STATES:
+                self.log("Suspending stack [%s]." % stack.name)
```
fghaas: Stack names get reused over time when learners reset labs. When examining this log message in the course of troubleshooting, it's obviously possible to use just the stack name in combination with `openstack stack list --deleted | grep $name` to then retrieve the uuid log of the deleted stack, but perhaps we want to instead log both the name and the uuid here?

hastexo/jobs.py
```diff
@@ -90,37 +90,43 @@ def suspend_stack(self, stack):
...
+                    stack.status = SUSPEND_RETRY_STATE
+                else:
+                    stack.status = heat_status
 
+                stack.save(update_fields=["status"])
+        except Exception as e:
+            # Schedule for retry on any uncaught exception
+            stack.status = SUSPEND_RETRY_STATE
             stack.save(update_fields=["status"])
+            self.log("Error suspending stack [%s]: %s" % (stack.name, str(e)))
```
fghaas: Same as above, perhaps log both stack name and uuid here?

hastexo/jobs.py
arbrandes: Great question: I originally thought this would never happen.  However, on versions prior to 2.5.3, unhandled exceptions (such as HTTPExceptions from Heat), could result in a bogus state in the database, including (but not limited to) a supposedly deletable stack without a provider.  I did my best to not let this happen in 2.5.3, but on the off chance I didn't fix all the holes, it's best to avoid the problem altogether "from the other side".

hastexo/jobs.py
arbrandes: Good point, and easy enough to fix.  I'm adding it in.

arbrandes: Actually, because this is a catch-all exception, we don't store the UUID in the database, and the most likely exception to trigger this error will be a failure to GET from Heat, we don't have access to the stack UUID here.

hastexo/jobs.py
arbrandes: Addresse don fix-up 4c682b4

hastexo/jobs.py
fghaas: Gotcha, works for me.

hastexo/jobs.py
fghaas: Makes sense, thanks!

hastexo/jobs.py
fghaas: Looks good to me; please squash — thanks!

PR #87, opened 2019-02-06
Python 3 support

fghaas:
It looks as though by tracking the `master` branch of edx/xblock-sdk, we can finally get the test harness to run in Python 3, so I took a stab at eliminating what's left of Python 2-only code.

- Replace `dict.iteritems()` and `dict.iterkeys()`
- Import Paramiko SSH exceptions, rather than referencing them by fully qualified class name
- Add Python 3.6 to the test matrix and require it to pass (but not 3.5 — see the relevant commit message for an explanation)

### DIFF
--- a/hastexo/hastexo.py
+++ b/hastexo/hastexo.py
@@ -396,7 +396,7 @@ class HastexoXBlock(XBlock,
                 provider_name = "default"
             else:
                 try:
-                    provider_name = configured_providers.iterkeys().next()
+                    provider_name = next(iter(configured_providers))
                 except StopIteration:
                     pass
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ language: python
 python:
   - "2.7"
   - "3.5"
+  - "3.6"
 
 before_install:
   - pip install codecov
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,3 +1,9 @@
+Unreleased
+---------------------------
+
+* [Enhancement] Track XBlock-SDK master branch in test matrix
+* [Bug fix] Python 3.6 compatibility
+
 Version 2.5.6 (2018-02-07)
 ---------------------------
 
--- a/hastexo/tasks.py
+++ b/hastexo/tasks.py
@@ -12,6 +12,9 @@ from celery.exceptions import SoftTimeLimitExceeded
 from heatclient.exc import HTTPException, HTTPNotFound
 from keystoneauth1.exceptions.http import HttpError
 from io import StringIO
+from paramiko.ssh_exception import (AuthenticationException,
+                                    SSHException,
+                                    NoValidConnectionsError)
 
 from .models import Stack
 from .heat import HeatWrapper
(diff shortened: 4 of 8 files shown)


hastexo/hastexo.py
```diff
@@ -396,7 +396,7 @@ def launch_stack(self, stack, settings, reset=False):
                 provider_name = "default"
             else:
                 try:
-                    provider_name = configured_providers.iterkeys().next()
+                    provider_name = next(iter(configured_providers))
```
fghaas: @arbrandes This is the line that I'm least sure of in the whole commit. Does this _actually_ iterate? (Also, did it before?)

`iter(d)` is a valid shorthand for `iter(d.keys())`, that part I did check.

fghaas:
GitHub is being silly again here, ordering the commits by (ostensible) date rather than decendance. I rebased and reordered the commits a couple of times; please use `git log -p` for the proper order.

hastexo/hastexo.py
arbrandes: This is just pythonese for "get me the first entry".  The change works just as well, AFAICT.

fghaas:
Hang on, let me rebase on current master and force-push so we have a clean commit history.

fghaas:
@arbrandes, anything you want to do about Python 3.5 support? Here's what breaks; it's always the same error:

```
ERROR: test_get_check_status (tests.unit.test_hastexo.TestHastexoXBlock)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/travis/build/hastexo/hastexo-xblock/tests/unit/test_hastexo.py", line 589, in test_get_check_status
    result = self.call_handler("get_check_status", {})
  File "/home/travis/build/hastexo/hastexo-xblock/tests/unit/test_hastexo.py", line 136, in call_handler
    make_request(data, method=method))
  File "/home/travis/build/hastexo/hastexo-xblock/.tox/py35-xblock10/lib/python3.5/site-packages/xblock/mixins.py", line 89, in handle
    return self.runtime.handle(self, handler_name, request, suffix)
  File "/home/travis/build/hastexo/hastexo-xblock/.tox/py35-xblock10/lib/python3.5/site-packages/xblock/runtime.py", line 1029, in handle
    results = handler(request, suffix)
  File "/home/travis/build/hastexo/hastexo-xblock/.tox/py35-xblock10/lib/python3.5/site-packages/xblock/mixins.py", line 64, in wrapper
    request_json = json.loads(request.body)
  File "/opt/python/3.5.6/lib/python3.5/json/__init__.py", line 312, in loads
    s.__class__.__name__))
TypeError: the JSON object must be str, not 'bytes'
```
I ask because Ubuntu Xenial comes with 3.5. Bionic has 3.6; it may well be that before we get to deploy a fully Python 3-enabled Open edX we'd actually be deploying on Bionic anyway.

arbrandes:
I'm not sure we _can_ do anything about it.  It seems like the problem is with the xblock SDK.  We might be able to work around it, but until the SDK officially supports Python3, would we even want to?

fghaas:
I actually think what's causing the error is the `data` that's being passed in, in our own `make_request()`. In 2.7 that would be a string, in 3.x that's a byte sequence. Which breaks in 3.5 and was then un-broken in 3.6.

fghaas:
@arbrandes This one was approved a while back, but was never merged. It's rebased on current master now; can you merge this please so we at least can check off our own Python 3 issues that we have in current master (regardless of how far along Open edX itself is)?

PR #89, opened 2019-02-28
Progress check hints via stderr

arbrandes:
Allow course authors to specify progress check hints by outputting
arbitrary messages on a script's stderr when it fails.

### DIFF
--- a/README.md
+++ b/README.md
@@ -393,7 +393,10 @@ You can also use the following nested XML options:
 * `tests`: A list of test scripts.  The contents of each element will be run
   verbatim a a script in the user's lab environment, when they click the "Check
   Progress" button.  As such, each script should define an interpreter via the
-  "shebang" convention.
+  "shebang" convention.  If any scripts fail with a retval greater than 0, the
+  learner gets a partial score for this instance of the XBlock.  In this case,
+  the `stderr` of failed scripts will be displayed to the learner as a list of
+  hints on how to proceed.
 
 For example, in XML:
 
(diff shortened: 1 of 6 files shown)


hastexo/public/js/main.js
```diff
@@ -444,6 +444,15 @@ function HastexoXBlock(runtime, element, configuration) {
                     dialog.find('input.ok').one('click', function() {
                         $.dialog.close();
                     });
+                    var hints_title = dialog.find('.hints_title').hide();
+                    var hints = dialog.find('.hints').empty().hide();
+                    if (data.errors.length > 0) {
+                        $.each(data.errors, function(i, error) {
+                            hints.append($('<li>', {text: error}));
```
fghaas: You should probably consider multi-line output on stderr here. I suppose you can either split the `error` content at newlines here and replace those with `<br/>`, _or_ (to be continued below)

hastexo/tasks.py
```diff
@@ -696,18 +697,23 @@ def run_tests(self, ssh):
 
             # Make it executable and run it.
             sftp.chmod(script, 0o775)
-            _, stdout, _ = ssh.exec_command(script)
+            _, stdout, stderr = ssh.exec_command(script)
             retval = stdout.channel.recv_exit_status()
             if retval == 0:
                 score += 1
+            else:
+                error = stderr.read()
```
fghaas: (continued from above) ... or make this `stderr.read.replace('\n', '<br/>')`. Not sure what's the better approach, to be honest.

tests/unit/test_tasks.py
```diff
@@ -989,5 +991,6 @@ def test_check_student_progress(self):
 
         # Assertions
         self.assertEqual(res["status"], "CHECK_PROGRESS_COMPLETE")
-        self.assertEqual(res["pass"], 2)
+        self.assertEqual(res["pass"], 1)
         self.assertEqual(res["total"], 3)
+        self.assertEqual(res["errors"], ["stderr message", "stderr message"])
```
fghaas: This would need another test for multi-line stderr output, as well.

README.md
```diff
@@ -421,27 +424,39 @@ For example, in XML:
...
   </hastexo>
 </vertical>
 ```
 
 **Important**: Do this only *once per section*. Defining it more that once
 per section is not supported.
 
+Note on tests: as seen in the above example, it is recommended to wrap them all
+in `<![CDATA[..]]>` tags.  This avoids XML parsing errors when special
+characters are encoutered, such as the `&` used to output to stderr in bash.
```
fghaas: Typo, should be "encountered".

Also, factually misleading: `&` sends a process to the background. I presume you meant the `&` in `>&2` but you'll have to reference `>&2` in full here.

hastexo/public/js/main.js
arbrandes: This is the better place to do it, I believe.  It's a display concern, so better to handle it in the code that does the displaying.

fghaas: Yep, I don't have a preference either way. Fine with either.

arbrandes:
Comments addressed in https://github.com/hastexo/hastexo-xblock/pull/89/commits/a28df67bae9ddfa486845fe37f1009203b5f0b32.

README.md
```diff
@@ -455,7 +455,7 @@ per section is not supported.
 
 Note on tests: as seen in the above example, it is recommended to wrap them all
 in `<![CDATA[..]]>` tags.  This avoids XML parsing errors when special
-characters are encoutered, such as the `&` used to output to stderr in bash.
+characters are encoutered, such as the `>&2` used to output to stderr in bash.
```
fghaas: Still missing a fix for the "encoutered" typo.

arbrandes: Ug, forgot about that one.  Fixed and squashed in https://github.com/hastexo/hastexo-xblock/pull/89/commits/1c8b6eab8a50e8bb8b5c53be7dc538c1328cce52.

PR #91, opened 2019-04-02
Docker support for guacamole

arbrandes:


fghaas:
I'm afraid neither my Docker nor my Guacamole proficiency is sufficient here for a proper review, but generally this looks good to me. I suggest we merge this.

arbrandes:
Ok, merging!

PR #93, opened 2019-05-31
Google Cloud Platform support

arbrandes:


arbrandes:
> This can probably be squashed into one of the prior commits?

Yeah.  The latter two commits, actually.  I just squashed them.

fghaas:
Huh. Travis not reporting is weird. Can you try restarting the CI build?

tests/unit/test_gcloud.py
```diff
@@ -0,0 +1,111 @@
...
+            "gc_token_uri": "",
+            "gc_auth_provider_x509_cert_url": "",
+            "gc_client_x509_cert_url": ""
+        }
+        patchers = {
+            "service_account": patch("hastexo.gcloud.service_account"),
+            "build": patch("hastexo.gcloud.build")
+        }
+        self.mocks = {}
+        for mock_name, patcher in patchers.iteritems():
```
fghaas: `iteritems()` is Python 2 only, will break in Python 3. Use `items()`.

hastexo/provider.py
```diff
@@ -1,20 +1,27 @@
 import time
-
+import base64
+import binascii
+import paramiko
+import random
+import string
+import yaml
+
+from StringIO import StringIO
```
fghaas: This only works in Python 2. Instead, use:

```python
try:
  from io import StringIO
except ImportError:
  from StringIO import StringIO
```

arbrandes: Good catch!  Changes addressed in 765ad43.

arbrandes:
@fghaas, yeah, I had fallen into the (bad) habit of not checking py36 locally.  Bunch of failures and at least one backward compatibility problem.  Working on it.

fghaas:
Excellent! Almost there; just one more tiny nit: when running `tox` locally myself, I just realized that we get some complaints from `flake8` if we're running under py3; these come from `xblock-utils` that get pulled into `src`. So, we'll have to add `src` to the [default flake8 `exclude` list](http://flake8.pycqa.org/en/latest/manpage.html), like so:

```diff
diff --git a/tox.ini b/tox.ini
index b5c4756..f865027 100644
--- a/tox.ini
+++ b/tox.ini
@@ -9,6 +9,7 @@ python =
 
 [flake8]
 ignore = E124,W504
+exclude = .svn,CVS,.bzr,.hg,.git,__pycache__,.tox,.eggs,*.egg,src
 
 [testenv]
 deps =
```
... and then we're squeaky clean again.

arbrandes:
Nope!  Merging, and thanks for the review!

PR #96, opened 2019-07-25
Reaper zombie destroyer

arbrandes:
It so happens that the Heat API, when under yet-undetermined but clearly adverse conditions (usually after long-running, failed launch attempts), will report a 404 for a stack that is actually still there.  This causes the XBlock state machine to remember that stack as deleted, with the undesireable effect that if the owning user never tries to launch that stack again, it will never be considered for reaping.

To avoid having such useless stacks lying around accruing running charges, this implements a catch-all phase to the reaper run.  This phase will list all stacks on each configured provider and compare each stack to its corresponding state in the database.  If the stack is deemed to be a zombie (i.e., it is known to be dead, but clearly isn't), it is immediately destroyed.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,4 +1,4 @@
-Unreleased
+Version 3.1.0 (2019-07-26)
 ---------------------------
 
 * [BACKWARD INCOMPATIBLE FOR GCP LABS ONLY] Encode GCP stack names
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,3 +1,11 @@
+Unreleased
+---------------------------
+
+* [BACKWARD INCOMPATIBLE FOR GCP LABS ONLY] Encode GCP stack names
+* [Enhancement] Provider stack listing
+* [Enhancement] Reaper zombie destroyer
+* [Enhancement] Add Python 3.7 test target
+
 Version 3.0.1 (2019-07-23)
 ---------------------------
 
--- a/tox.ini
+++ b/tox.ini
@@ -1,11 +1,12 @@
 [tox]
-envlist = py{27,35,36}-xblock{10,11,12},flake8
+envlist = py{27,35,36,37}-xblock{10,11,12},flake8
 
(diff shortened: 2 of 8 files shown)


tox.ini
```diff
@@ -1,11 +1,12 @@
 [tox]
-envlist = py{27,35,36}-xblock{10,11,12},flake8
+envlist = py{27,35,36,37}-xblock{10,11,12},flake8
```
fghaas: I obviously don't object to adding 3.7 to the test matrix at all, but just curious about the rationale. As far as I can tell the only Ubuntu platform that deploys 3.7 is disco, which is not an LTS and hence, I think, is not a valid deployment platform for Open edX... am I missing something?

Changelog.md
```diff
@@ -1,3 +1,11 @@
+Unreleased
+---------------------------
+
+* [BACKWARD INCOMPATIBLE] Encode/decode GCP stack names
```
fghaas: Might perhaps want to be a bit more specific here: if I understand correctly, the compatibility breaking change _only_ affects GCP labs, whereas OpenStack labs should be unaffected. Correct?

tox.ini
arbrandes: I could fall back on the future-proofness argument, but the reality is that I put this in because it was the easiest way for me to get a python3 tox run on my Disco box. :grin: 

If you know of a simple way of installing different python versions on bare metal just for testing purposes, please share.  (I know I could just use containers, but I'm looking for a way to integrate test runs into my editor, and that would make it harder.)

Changelog.md
arbrandes: Correct, it only affects GCP labs, and since there are no known runs of it in production, 3.0.0 was only just released, and there's no non-goldbergian way to make this backward compatible, I figured "what the heck".

arbrandes: But yes, this deserves clarification.  Ammending.

PR #85, opened 2019-01-30
[WIP] v2.5.5

arbrandes:

PR #99, opened 2019-08-08
tasks.py: continue on specific EnvironmentErrors during SSH connection

fghaas:
When attempting to connect to a stack with Paramiko, we might not yet have full connectivity to SSH; in particular, a virtual router may not be fully operational yet. Under such circumstances, we might encounter a `socket.error`, such as error 104 `Connection reset by peer`.

Previously, such a `socket.error` would cause a stack launch (or stack resumption) to needlessly fail.

Add `socket.error` to the list of exceptions that cause Paramiko to retry, rather than give up outright.

Please note: this is me cargo-culting #85, and I do not know why it's currently breaking the test, so I'd appreciate a suggestion for a fix.

### DIFF
--- a/hastexo/tasks.py
+++ b/hastexo/tasks.py
@@ -1,3 +1,4 @@
+import errno
 import time
 import os
 import uuid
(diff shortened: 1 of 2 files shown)


fghaas:
This might also be too trivial. As of Python 3.3, `socket.error` [is an alias](https://docs.python.org/3/library/socket.html#socket.error) of `OSError`, so this probably retries too broadly. We might want to filter specifically `OSError` with the following `errno` attributes:

```
#define EAGAIN          11      /* Try again */
#define ENETDOWN        100     /* Network is down */
#define ENETUNREACH     101     /* Network is unreachable */
#define ENETRESET       102     /* Network dropped connection because of reset */
#define ECONNABORTED    103     /* Software caused connection abort */
#define ECONNRESET      104     /* Connection reset by peer */
#define ENOTCONN        107     /* Transport endpoint is not connected */
#define ECONNREFUSED    111     /* Connection refused */
#define EHOSTDOWN       112     /* Host is down */
#define EHOSTUNREACH    113     /* No route to host */
```

fghaas:
Scratch `ECONNREFUSED` and `EHOSTUNREACH` from that list; it looks like Paramiko [treats those specially](https://github.com/paramiko/paramiko/blame/ccf48a12f35db3c4becbb4b2e695a445da983f24/paramiko/client.py#L355) already.

arbrandes:
No idea why these tests are failing.  Gonna try and reproduce locally.

Otherwise, I have no objections to the patch except that, as you suggest above, we should be specific about which ERRNOs we retry on.

arbrandes:
@fghaas, check [my PR to your PR](https://github.com/fghaas/hastexo-xblock/pull/1).

fghaas:
@arbrandes, thanks for fixing that. I've added a bit more logging; see my most recent commit. Let me know if you're good with that, and if you are I'll squash and force-push.

arbrandes:
@fghaas, no objections, looks good!

fghaas:
@arbrandes: Ack, thanks! I've squashed everything into one commit and added a `Co-Developed-by:` tag to the commit message, per precedent set by [the corresponding Linux kernel convention](https://www.kernel.org/doc/html/v4.17/process/submitting-patches.html#when-to-use-acked-by-cc-and-co-developed-by). Let me know if you're good with that.

arbrandes:
I'm perfectly good with this.

PR #102, opened 2019-08-14
v3.2.1

arbrandes:
Contains two bug fixes.


hastexo/migrations/0005_auto_20190811_1555.py
```diff
@@ -36,7 +36,7 @@ class Migration(migrations.Migration):
         migrations.AddField(
             model_name='stacklog',
             name='hook_events',
-            field=jsonfield.fields.JSONField(default=None),
+            field=jsonfield.fields.JSONField(default=b'null'),
```
fghaas: I'm a bit puzzled as to why this is a binary string, but I guess this is from `manage.py makemigrations`?

arbrandes: Correct.  Straight from `makemigrations`.

arbrandes:
> ~Looks good to me, one minor question.~ Edit: sadly, this breaks on Python 3.

Yeah, nice catch on the binary string.  That's what fails on Python 3.  Fixed the migration manually, once more.

arbrandes:
Wat.  I tested py27 locally and it worked just fine.

PR #105, opened 2019-10-16
v3.3.0

arbrandes:


arbrandes:
@fghaas, objections?  I've tested this on an ironwood.2 production deployment and it works fine.

requirements/base.txt
```diff
@@ -1,17 +1,13 @@
...
-python-heatclient==1.6.1
-python-novaclient==7.1.2
-google-api-python-client==1.7.7
-paramiko==2.1.6
-celery==3.1.18
-apscheduler==3.5.1
-jsonfield==2.0.2
-six==1.11.0
--e git+https://github.com/edx/xblock-utils.git@v1.0.5#egg=xblock-utils==v1.0.5
+apscheduler>=3.5.1,<3.6
```
fghaas: No real objection here, just curious why all of these do specify maximum versions.

requirements/test.txt
```diff
@@ -1,6 +1,10 @@
 -r base.txt
 django>=1.11,<2.0
+celery>=3.1.25,<4.0
 -e git://github.com/edx/xblock-sdk.git@master#egg=xblock-sdk==master
+-e git+https://github.com/edx/xblock-utils.git@v1.0.5#egg=xblock-utils==v1.0.5
```
fghaas: Any particular reason for not using 1.2.2 here, and for installing from GitHub as opposed to [PyPI](https://pypi.org/project/xblock-utils/)?

fghaas:
@arbrandes: Just re-upping this, so that it doesn't fall off the table. :)

requirements/base.txt
arbrandes: The idea is to have a relaxed set of requirements in setup.py for maximum runtime flexibility (an attempt at reducing last-minute fixes when some pip dependency gets upgraded), but a more reproducible environment in requirements.txt - who knows if a future Heat version will be incompatible with the current API.  A point release freeze would work, but then we wouldn't get the odd security/bug fix without manual work.  Hence, a middle ground with maximum versions.

requirements/test.txt
arbrandes: The only reason was that it didn't use to be on PyPI, and I never checked to see if they pushed it up.  Fixed on latest push.

arbrandes:
Huh.  This now fails just on Python 3.6.  And also just on Travis: locally, with an Anaconda-installed py36, it works fine.  I'll mess with the test requirements and push attempts to fix it.

PR #106, opened 2019-12-17
v3.4.0

arbrandes:


fake/student/models.py
```diff
@@ -0,0 +1,15 @@
+from django.contrib.auth.models import User
+from django.db import models
+
+
+class AnonymousUserId(models.Model):
+    """
+    Paired down from:
```
fghaas: Nano-nit: typo, should be "pared down".

arbrandes: Whoops, excellent catch.  Thanks!

PR #107, opened 2020-01-07
v3.4.1

arbrandes:


fghaas:
I know so little about `JSONField` that I can't reasonably comment on the merits here, but assuming this is the latest of what you have deployed it appears to work fine — so I'm good with this being merged. Thanks!

arbrandes:
I'm amend this with a fix for the dropdowns shortly.

PR #108, opened 2020-01-17
test: Require django>=1.11.27 (CVE-2019-19844)

fghaas:
Django releases earlier than 1.11.27 are vulnerable to account
hijack as detailed in CVE-2019-19844. Since Open edX currently
requires Django<1.12, update the test requirements to allow any
1.11.z version for z>=27.


arbrandes:
Merged as part of https://github.com/hastexo/hastexo-xblock/pull/110

PR #109, opened 2020-01-17
Update test matrix

fghaas:
* Bump the XBlock 1.2 version we test against to the latest, 1.2.9
* Add Python 3.8


arbrandes:
Merged in https://github.com/hastexo/hastexo-xblock/pull/110

PR #111, opened 2020-03-02
Allow nested blocks

arbrandes:
Let the course author include HTML, PDF, or video blocks as children.

Motivation: to allow lab instructions be grouped together with the correct instance, when served as part of a content library.

TODO:
- [x] [Hack library to allow nested blocks](https://github.com/arbrandes/edx-platform/commit/ca0f8e0824d7125eb118cb0b837d876f3ed41eea)
- [x] Tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,7 +77,20 @@ To deploy the hastexo XBlock:
     ],
     ```
 
-4. This xblock uses a Django model to synchronize stack information across
+4. If you're going to use it in a content library, also add it to the
+   `ADVANCED_PROBLEM_TYPES` of your Studio environment, by editing
+   `/edx/app/edxapp/cms.env.json` and adding:
+
+    ```
+    "ADVANCED_PROBLEM_TYPES": [
+        {
+            "boilerplate_name": null,
+            "component": "hastexo"
+        }
+    ],
+    ```
+
+5. This xblock uses a Django model to synchronize stack information across
    instances.  Migrate the `edxapp` database so the `hastexo_stack` table is
    created:
 
@@ -85,7 +98,7 @@ To deploy the hastexo XBlock:
    $ sudo /edx/bin/edxapp-migrate-lms
    ```
 
-5. Add configuration to `XBLOCK_SETTINGS` on `/edx/app/edxapp/lms.env.json`:
+6. Add configuration to `XBLOCK_SETTINGS` on `/edx/app/edxapp/lms.env.json`:
 
     ```
     "XBLOCK_SETTINGS": {
@@ -144,7 +157,7 @@ To deploy the hastexo XBlock:
     }
     ```
 
-6. Now install the Guacamole web app and stack supervisor scripts by cloning
+7. Now install the Guacamole web app and stack supervisor scripts by cloning
    the `hastexo_xblock` fork of edx/configuration and assigning that role to
    the machine:
 
@@ -154,7 +167,7 @@ To deploy the hastexo XBlock:
     $ ansible-playbook -c local -i "localhost," run_role.yml -e role=hastexo_xblock
     ```
 
-7. At this point restart edxapp, its workers, and make sure the stack jobs are
+8. At this point restart edxapp, its workers, and make sure the stack jobs are
    running:
 
     ```
@@ -164,7 +177,7 @@ To deploy the hastexo XBlock:
     sudo /edx/bin/supervisorctl start reaper:
     ```
 
-8. Finally, in your course, go to the advanced settings and add the hastexo
+9. Finally, in your course, go to the advanced settings and add the hastexo
    module to the "Advanced Module List" like so:
 
    ```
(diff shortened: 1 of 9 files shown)


hastexo/hastexo.py
```diff
@@ -232,10 +240,32 @@ def parse_xml(cls, node, runtime, keys, id_generator):
...
+            additional_blocks.append(_spec)
+        except ImportError:
+            pass
+
+        try:
+            from pdf import pdfXBlock
+            _spec = NestedXBlockSpec(pdfXBlock, category="pdf", label=u"PDF")
+            additional_blocks.append(_spec)
+        except ImportError:
+            pass
```
fghaas: Same consideration applies as above for `VideoDescriptor`, though I do realize that the PDF XBlock is an entirely optional add-on component. So, perhaps this should be logged with `logging.info()` or `logging.debug()` instead.

hastexo/hastexo.py
```diff
@@ -232,10 +240,32 @@ def parse_xml(cls, node, runtime, keys, id_generator):
...
+        """
+        additional_blocks = []
+        try:
+            from xmodule.video_module.video_module import VideoDescriptor
+            _spec = NestedXBlockSpec(
+                VideoDescriptor, category="video", label=u"Video"
+            )
+            additional_blocks.append(_spec)
+        except ImportError:
+            pass
```
fghaas: In this case I'm curious _how_ an import from `xmodule.video_module.video_module` would fail unless edx-platform is really really badly misconfigured. Maybe log this at the warning level? As in:
```python
except ImportError:
  logging.warning("Unable to import VideoDescriptor", exc_info=True)
```

hastexo/hastexo.py
```diff
@@ -232,10 +240,32 @@ def parse_xml(cls, node, runtime, keys, id_generator):
 
         return block
 
-    def author_view(self, context=None):
```
fghaas: Not quite sure why this method was dropped? Does it no longer apply and if it doesn't, what can the author expect to render in Studio, here?

hastexo/hastexo.py
arbrandes: This is in case the XBlock is running from the workbench.  I realize we don't really support this anyway, but it is used as an example in `olx-utils`, and I figured it was harmless to keep the suggested behavior.

hastexo/hastexo.py
arbrandes: `xblock-utils` uses this view to render the children.  I think overriding it just for the somewhat obvious warning that the terminal doesn't work in Studio is not worth the added maintenance overhead of having to keep the overriden function in sync with `xblock-utils`.

hastexo/hastexo.py
```diff
@@ -43,14 +47,18 @@ def __init__(self, error_msg):
...
     Provides lab environments and an SSH connection to them.
 
     """
+    CATEGORY = "hastexo"
+    STUDIO_LABEL = "hastexo XBlock"
+
     # Settings with defaults.
     display_name = String(
-        default="Lab",
+        default="hastexo XBlock",
```
fghaas: Does this mean that all courses using this component would need to be updated so they override this new default to keep saying `Lab`?

requirements/test.txt
```diff
@@ -1,16 +1,22 @@
 -r base.txt
+
+# These package versions must be kept in sync with edx-platform as much as possible.
 django>=1.11.27,<1.12
 celery>=3.1.25,<4.0
+xblock-utils>=1.2.0,<1.3
+jsonfield==2.0.2
```
fghaas: Since the _package_ itself does depend on `jsonfield`, and not just the unit/integration tests, this also requires an entry in the `install_requires` list in `setup.py`. Presumably like:

```python
    install_requires=[
        'apscheduler',
        'google-api-python-client',
        'jsonfield>=2.0.2,<3.0',
        'paramiko',
        'python-heatclient',
        'python-keystoneclient',
        'python-novaclient',
    ],
```

hastexo/static/html/main.html
```diff
@@ -1,4 +1,5 @@
-<div class="hastexo_block">
+{{child_content|safe}}
+<div class="hastexblock">
```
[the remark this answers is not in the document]
arbrandes: No.  The only reference to this class is in main.js.  I had to change it because JQuery refused to find it, due to the underscore.  More likely than not a JQuery bug, but I figured changing this was the more efficient fix.

hastexo/hastexo.py
arbrandes: Yes, but note that this is only really visible in Studio.  In the LMS, the display name of individual blocks is invisible.

I changed it because saying "Lab" is less than descriptive, specially when picking from the list of advanced problems in Studio.

requirements/test.txt
arbrandes: No, we're not doing that anymore.  Package dependencies that come via edX are not to be specified explicitly, so as to avoid runtime conflicts when edX pops an upgrade.

arbrandes: (Except on test requirements, of course.  Instead of depending on edx-platform itself, it is much more efficient to duplicate the dependencies edx-platform introduces and on which the XBlock depends.)

README.md
```diff
@@ -565,6 +575,35 @@ In order to add the hastexo Xblock through Studio, open the unit where you want
...
+### Using the hastexo XBlock in a content library
+
+This XBlock is usable in [content libraries](https://edx.readthedocs.io/projects/open-edx-building-and-running-a-course/en/latest/course_components/libraries.html).
+It supports adding lab instructions as child blocks, so that when the block is
+randomized, the instructions are bundled together with it.
+
+To add the XBlock to the library via Studio, make sure it is configured as one
+of the `ADVANCED_PROBLEM_TYPES` in `cms.env.json`, then select it as such when
+adding content to your library.  (Note: as of Open edX Ironwood, the ability to
+do so requires running a patched version of `edx-platform`.)
```
fghaas: Can we have a link to the patched branch here, please?

hastexo/hastexo.py
fghaas: OK, makes sense.

hastexo/static/html/main.html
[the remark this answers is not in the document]
fghaas: OK, fair enough. :)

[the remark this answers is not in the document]
fghaas: It would be good if you could mention this in the commit message though. Or in an inline comment, lest a future adopter change this to something that _does_ involve an underscore again, mysteriously breaking things.

arbrandes:
> Just checking: we're confident that _existing_ instances of the hastexblock, i.e. ones without nested blocks, will continue to work just fine, and unchanged?

Yes.  This is really just an addition.  It doesn't remove or replace any functionality, or change any explicit or implicit contracts except for the "hastexblock" class in the DOM, a class which is only used internally by the XBlock JS code.

README.md
arbrandes: Done.

arbrandes:
I give up trying to get Travis to complete.

fghaas:
Merged after a complete and successful local run of the test suite. Thank you!

PR #112, opened 2020-03-12
Retry database status update

fghaas:
@arbrandes, maybe you still want to give this a look. It's a WIP PR for doing database update retries from tasks, and it would (currently) _only_ apply to the update at the very end of the stack launch.

The idea is, for that specific update, to retry on database errors until it either goes away, or we hit the stack timeout (obviously with a sleep in between).

### DIFF
--- a/hastexo/tasks.py
+++ b/hastexo/tasks.py
@@ -3,7 +3,8 @@ import os
 import traceback
 import socket
 
-from django.db import transaction
+from django.db import transaction, close_old_connections
+from django.db.utils import OperationalError
 from celery import Task
 from celery.utils.log import get_task_logger
 from celery.exceptions import SoftTimeLimitExceeded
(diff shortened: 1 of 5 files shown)


hastexo/tasks.py
```diff
@@ -165,14 +167,51 @@ def run(self, **kwargs):
...
+                self.update_stack(data)
+                break
+            except SoftTimeLimitExceeded:
+                # If we've hit the timeout, something went massively
+                # wrong, repeatedly. Log an error, and throw the
+                # timeout up the stack.
+                logger.error("Timeout updating database status for "
+                             "[%s]." % self.stack_name)
+                raise
+            except DatabaseError:
```
fghaas: Retry just on `django.db.utils.OperationalError` here, maybe?

hastexo/tasks.py
```diff
@@ -165,14 +167,51 @@ def run(self, **kwargs):
...
+        # discouraged.
+        #
+        # Reference:
+        # https://docs.djangoproject.com/en/1.11/topics/db/transactions/
+        while True:
+            try:
+                # Try updating the stack. If it succeeds, we're done.
+                self.update_stack(data)
+                break
+            except SoftTimeLimitExceeded:
```
fghaas: Not sure if we even want to catch this here. We can't do cleanup, which is what we normally do when we hit `SoftTimeLimitExceeded`, so we may just want to throw this up the stack.

hastexo/tasks.py
arbrandes: I'm obviously leery of retrying things automatically (otherwise I'd have implemented something like this sooner), and particularly leery of infinite retries.  That said, it does make sense to retry - but only sometimes.  One way of coding that in is precisely to restrict retries to known problems.  So yes, I'd be as specific as possible, here.

hastexo/tasks.py
arbrandes: It's still better to be able to try doing something, when all has gone up the creek.  I'd leave it in.

arbrandes:
I think the proposed change here is sound, but my gut feeling is that it's trying too hard.  I'd restrict the retries to more specific database errors, and I would not retry until the timeout without at least some sort of incremental sleep back-off.  If the DB goes away, for example, you may end up pegging app servers and making it more difficult to debug the issue.

hastexo/tasks.py
fghaas: Right, I understand that. But the reasoning that this being a Celery task, the retry isn't actually infinite, and the retry happens outside the atomic block and uses a sleep, so shouldn't be blocking things globally — that's not incorrect, on face value, is it?

hastexo/tasks.py
fghaas: Ack, thanks. Works for me.

fghaas:
> I think the proposed change here is sound, but my gut feeling is that it's trying too hard. I'd restrict the retries to more specific database errors, and I would not retry until the timeout without at least some sort of incremental sleep back-off. If the DB goes away, for example, you may end up pegging app servers and making it more difficult to debug the issue.

Right, so you're suggesting something like
* sleep for the configured sleep timeout times 1 (2^0), 
* on next iteration sleep for timeout times 2 (2^1),
* on next iteration sleep for timeout times 4 (2^2),
* on next iteration sleep for timeout times 8 (2^8),
* and so on, until we hit the timeout?

hastexo/tasks.py
fghaas: I've restricted this to just OperationalError in the 
9f9ac5a fixup (which I obviously intend to squash).

fghaas:
The exponential back-off is now in 9f9ac5a5e08205dafa5a7917c4121b2e0b5be348.

hastexo/tasks.py
arbrandes: > (...) the reasoning that this being a Celery task, the retry isn't actually infinite, and the retry happens outside the atomic block and uses a sleep, so shouldn't be blocking things globally — that's not incorrect, on face value, is it?

Nope, not incorrect at all.  Or, to avoid the double negative, it's perfectly correct, AFAICT.

arbrandes:
> * sleep for the configured sleep timeout times 1 (2^0),
> * on next iteration sleep for timeout times 2 (2^1),
> * on next iteration sleep for timeout times 4 (2^2),
> * on next iteration sleep for timeout times 8 (2^3),
> * and so on, until we hit the timeout?

Yes, exactly.  With maybe an upper limit on the top interval, so it doesn't wait for hours on the last step.

arbrandes:
It looks like just retrying is not enough.  I suggest trying to `close_old_connections()` after each failure, much like the suspender and reaper jobs [have to](https://github.com/hastexo/hastexo-xblock/blob/master/hastexo/jobs.py#L45).

fghaas:
> It looks like just retrying is not enough. I suggest trying to `close_old_connections()` after each failure, much like the suspender and reaper jobs [have to](https://github.com/hastexo/hastexo-xblock/blob/master/hastexo/jobs.py#L45).

Would you suggest that just in the `except OperationalError` clause, or in a `finally` block?

arbrandes:
> Would you suggest that just in the `except OperationalError` clause, or in a `finally` block?

The way you did it is how I would've suggested it. :)

fghaas:
@arbrandes, I just added the bits that would extend this retry logic to database updates on suspend and delete. As discussed elsewhere, I moved the stack update methods to the base class, and then made minor adjustments to the child classes as well, as the commit message on 6164154 explains.

Does this look OK to you?

arbrandes:
> Does this look OK to you?

Looks great!  No objections.

PR #113, opened 2020-03-19
Bump version: 3.4.2 → 3.5.0

arbrandes:


fghaas:
I thought we'd make this one 3.4.3, and #111 3.5.0. Or do you think we ought to bump the major for #111?

arbrandes:
> I thought we'd make this one 3.4.3, and #111 3.5.0. Or do you think we ought to bump the major for #111?

Either way is fine.  But I figured the retry was sufficiently different, and an enhancement at that, to warrant a minor bump (vs a patch bump).

fghaas:
I'll defer to your call. I guess it's debatable whether being persistent with database updates is a bug fix, or an enhancement. But yeah I put it into the changelog as an enhancement myself, so you're right. :)

arbrandes:
Merged.  I'll need you to upload to pypi, though, as I'm no longer an admin. :)

fghaas:
[Done](https://pypi.org/project/hastexo-xblock/#history), thank you!

PR #117, opened 2020-04-20
Retry get_provider_stack_count()

fghaas:
Rather than only retrying up to three times for updating the database
stack status, also apply that same logic to fetching the stack count
per provider.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,6 +1,7 @@
 Unreleased
 ---------------------------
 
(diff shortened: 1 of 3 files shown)


elenalind:
> Rather than only retrying up to three times for updating the database
> stack status, also apply that same logic to fetching the stack count
> per provider.

@fghaas I'd like to check this via git in terminal but I can't see this branch with `git branch -a` .

PR #129, opened 2020-07-30
Make terminal color scheme and font configurable

mrtmm:
Add configuration options to edit guacamole terminal color scheme
and font by defining the following variables via EDXAPP_XBLOCK_SETTINGS:
'terminal_color_scheme', 'terminal_font_name', 'terminal_font_size'.

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,6 +100,9 @@ To deploy the hastexo XBlock:
     "XBLOCK_SETTINGS": {
         "hastexo": {
             "terminal_url": "/hastexo-xblock/",
+            "terminal_color_scheme": "white-black",
+            "terminal_font_name": "monospace",
+            "terminal_font_size": 10,
             "launch_timeout": 900,
             "remote_exec_timeout": 300,
             "suspend_timeout": 120,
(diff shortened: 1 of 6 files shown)


hastexo/common.py
```diff
@@ -148,6 +148,9 @@
 
 DEFAULT_SETTINGS = {
     "terminal_url": "/hastexo-xblock/",
+    "terminal_color_scheme": "white-black",
+    "terminal_font": "monospace",
```
fghaas, line 152: Perhaps call this `terminal_font_name` to maintain parallelism with Guacamole options? (This would also require an update to the commit message.)

mrtmm: Good point, will do that.

README.md
```diff
@@ -202,6 +205,18 @@ This is a brief explanation of each:
   such as ":8080/hastexo-xblock/", for use in devstacks). (Default:
   `/hastexo-xblock/`)
 
+* `terminal_color_scheme`: Color scheme for the terminal window. Suitabale values
+  are described in [Guacamole Documentation](https://guacamole.apache.org/doc/gug/configuring-guacamole.html#ssh).
+  For example, `foreground: rgb:ff/ff/ff;background: rgb:00/00/00` as well as
```
fghaas, line 210: Suggest to replace "as well as" with "and", and remove the comma before "both".

PR #131, opened 2020-08-11
Add configuration option for instructions layout

mrtmm:
Add variable 'instructions_layout' to XBLOCK_SETTINGS that allows
positioning lab intructions either above, below, left or right
from the terminal window.

### DIFF
--- a/hastexo/public/js/main.js
+++ b/hastexo/public/js/main.js
@@ -14,6 +14,9 @@ function HastexoXBlock(runtime, element, configuration) {
     var dialog_container = undefined;
 
     var init = function() {
+        /* Construct the layout for instructions and terminal */
+        construct_layout();
+
         /* Set dialog container. */
         dialog_container = $(element).find('.hastexblock')[0];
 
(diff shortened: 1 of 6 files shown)


hastexo/public/js/main.js
```diff
@@ -270,6 +273,52 @@ function HastexoXBlock(runtime, element, configuration) {
         });
     };
 
+    var construct_layout = function() {
+        var instructions_layout = configuration.instructions_layout;
+
+        /* 'above' is the default layout and doesn't require any changes */
+        if (instructions_layout != 'above') {
+            var content = document.getElementsByClassName("vert-mod")[0];
```
fghaas, line 281: I *think* you can make this a little more readable by using a JQuery selector, like we do for selecting `dialog_container`.

hastexo/public/js/main.js
```diff
@@ -270,6 +273,52 @@ function HastexoXBlock(runtime, element, configuration) {
...
+
+            var parent_element = function(child_element) {
```
fghaas, line 283: This function can *probably* also be expressed with a JQuery selector.

hastexo/public/js/main.js
```diff
@@ -270,6 +273,52 @@ function HastexoXBlock(runtime, element, configuration) {
...
+                for (var div of content.children)
+                    if (div.contains(child_element)) {
+                        return div;
+                    };
+            };
+            var instructions = document.getElementsByClassName('lab_instructions')[0];
```
fghaas, line 289: Also, probably a case for JQuery.

hastexo/public/js/main.js
```diff
@@ -270,6 +273,52 @@ function HastexoXBlock(runtime, element, configuration) {
...
+
+            var terminal_parent = parent_element(document.getElementById('terminal'));
+            var instructions_parent = parent_element(instructions);
```
fghaas, line 292: Here, too.

hastexo/public/js/main.js
```diff
@@ -270,6 +273,52 @@ function HastexoXBlock(runtime, element, configuration) {
...
+
+            if (terminal_parent != 'undefined' && instructions_parent != 'undefined') {
```
fghaas, line 294: It looks like the condition in parentheses here should always be true. Do you perhaps want an `else` block here that writes a warning to the JavaScript console when it's false, to catch logic bugs?

hastexo/public/js/main.js
```diff
@@ -270,6 +273,48 @@ function HastexoXBlock(runtime, element, configuration) {
...
+                    /* if terminal is on the left side, move terminal buttons to the left as well */
+                    if (instructions_layout === 'right') {
+                        $(element).find('.buttons').css({'text-align': 'left'});
+                    };
+                };
+                if (instructions_layout === 'below') {
+                    $(instructions_parent).insertAfter($(terminal_parent));
+                };
+            } else {
+                console.warn('Unable to edit content layout, elements not found');
```
fghaas, line 313: "Edit" usually implies user interaction, perhaps use "modify" or "adjust" here.

mrtmm: Ack, done.

fghaas:
Looks good to me! Let me know in case you have any objections to me merging this.

mrtmm:
No objections, this is ok to merge.

PR #132, opened 2020-08-12
Minor fixes to accomodate instructions layout options

mrtmm:
Add margin to dialog-overlay to remove overlapping background and
shadow from the sides.
Calculate content height for side views dynamically to make sure
xblock fits the area.

### DIFF
--- a/hastexo/public/js/main.js
+++ b/hastexo/public/js/main.js
@@ -287,9 +287,11 @@ function HastexoXBlock(runtime, element, configuration) {
 
             if (terminal_parent != 'undefined' && instructions_parent != 'undefined') {
                 if (instructions_layout === 'left' || instructions_layout === 'right') {
-                    $(content).addClass('content-side-by-side');
                     $('.lab_instructions').addClass('instructions-side-view');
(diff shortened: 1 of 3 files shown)


hastexo/public/js/main.js
```diff
@@ -287,9 +287,11 @@ function HastexoXBlock(runtime, element, configuration) {
 
             if (terminal_parent != 'undefined' && instructions_parent != 'undefined') {
                 if (instructions_layout === 'left' || instructions_layout === 'right') {
-                    $(content).addClass('content-side-by-side');
                     $('.lab_instructions').addClass('instructions-side-view');
                     $('#container').addClass('terminal-side-view');
+                    $(content).addClass('content-side-by-side');
+                    /* Make sure the xblock fits to content area */
+                    $(content).height($('.hastexblock').height() + 20);
```
fghaas, line 294: Just for the benefit of my understanding: what dimension does `20` use here? Is that pixels?

mrtmm: Yes, pixels.

PR #133, opened 2020-10-21
Disable page auto-scroll when terminal gets focused

mrtmm:


fghaas (changes requested):
The functional change itself is perfectly fine. Can we please have a little more detail in the commit message, though? Such as, what "auto-scroll" means, why we consider it detrimental, that it affects Chromium and Safari users but not Firefox users, and also that the patch fixes the behavior on Chrome (and probably all Chromium-based browsers), but not on Safari?

mrtmm:
Sure, added a more informative commit message.

fghaas:
> Safari however, is auto-scrolling and this commit does not enable that.

This part of the commit message is still confusing. `s/enable/disable/` ?

mrtmm:
Yes, sorry that should be `disable`.

PR #134, opened 2020-10-28
Allow overriding 'delete_age' by defining an XBlock attribute

mrtmm:
This commit adds a 'delete_age' field to XBlock attributes. If set, it will override
the global setting.
When a lab gets suspended, a 'delete_by' timestamp will be added to it, which will be
calculated by adding the 'delete_age' amount of time to the moment of suspension.
The reaper will use the 'delete_by' timestamps to find and delete old lab stacks. The
'delete_by' value can additionally be overridden in the admin panel.


hastexo/admin.py
```diff
@@ -99,7 +100,7 @@ class StackAdmin(admin.ModelAdmin):
     search_fields = ("name", "course_id", "status", "provider")
     readonly_fields = ("name", student_email, "course_id", "run", "protocol",
                        "port", "ip", "launch_timestamp", "suspend_timestamp",
-                       "created_on", "error_msg",)
+                       "created_on", "error_msg", "delete_age",)
```
fghaas, line 103: Help me understand this, please. If `delete_age` is read-only in the admin interface and `delete_by` is writable, what are the semantics of setting `delete_by` and leaving `delete_age` unmodified?

hastexo/tasks.py
```diff
@@ -625,6 +626,11 @@ def run(self, **kwargs):
             'status': status,
         }
 
+        # Set the delete_by value based on the delete_age value
+        if not stack.delete_by and status != SUSPEND_FAILED:
```
fghaas, line 630: Wait, hold on. This doesn't look right.

First, I don't think it make sense to set `delete_by` only if it isn't already set. The idea is that a stack gets reaped _if it hasn't been resumed in some time,_ so setting the `delete_by` state only on the first suspend, rather than updating it on every suspend, defeats that purpose.

Second, I think we want to set this _even if_ `status == SUSPEND_FAILED` because otherwise a stack might fail to suspend, and subsequently linger forever as a result. 

Or am I missing something?

Changelog.md
```diff
@@ -1,3 +1,8 @@
+Unreleased
+---------------------------
+
+* [Enhancement] Enable overriding 'delete_age' via XBlock attribute
```
fghaas, line 4: You'll need to elaborate more here. In particular, you'll want to expand on the days (in configuration settings) vs. seconds (in XBlock settings) discrepancy, why it exists, and how it will go away in the future.

README.md
```diff
@@ -508,6 +508,9 @@ The following are optional:
 * `launch_timeout`: How long to wait for a stack to be launched, in seconds.
   If unset, the global timeout will be used.
 
+* `delete_age`: Delete stacks that haven't been resumed in this many seconds.
+  Overrides the globally defined setting.
```
fghaas, line 512: Here, you probably also want to explain _why_ this is expressed in seconds, not days.

hastexo/hastexo.py
```diff
@@ -91,6 +91,12 @@ class HastexoXBlock(XBlock,
         scope=Scope.settings,
         help="The relative path to an uploaded executable script. "
              "For example, \"hot_script.sh\".")
+    delete_age = Integer(
```
fghaas, line 94: So, down the road we want to support not only a plain number that we interpret as seconds, but also suffixes like `3d` for three days and `1w` for one week.

Thus, I am guessing that you want this to be a String, which for now you then internally *cast* to an integer. Does that make sense?

fghaas (changes requested):
Thanks, I've gone through it now and I do have a few questions. :)

hastexo/admin.py
mrtmm: Well, I saw two options for this:
1.` delete_age` get set each time the student starts or resumes their lab. It will look for the XBlock attribute value if that is `None`, use the global settings value. And with this approach, if the value is set in the admin it will simply be overridden when the student resumes, so having the option to edit it would be just pointless in this case. But this gives an option to edit the `delete_age` even mid-course for all students. The `delete_by` value can be set in the admin to optionally just throw away labs even before students resume (thinking that this would be a rare case).

2. `delete_age` gets set once when the student first launches the stack. It will look for the XBlock attribute value if that is `None`, use the global settings value. When the student resumes, if `delete_age` has a value, don't change it. So if we went and edited the value in the admin, the value will now persist. But we cannot make the change later via the XBlock attribute that would apply for all. If we want to change the delete_age for a single stack, we can do this here but if we would want to change it for all students, I am not sure how easy/comfortable would be doing that in the admin.

So I went with the first option and made the `delete_age` read-only, simply because the change would not take affect.

hastexo/tasks.py
mrtmm: Right, I'll need to trace my thoughts on this, I believe you are right. We should just set the `delete_by` value here I think.

hastexo/hastexo.py
mrtmm: Yes, makes sense.

fghaas (changes requested):
This looks good to me now, except the commit message still needs an update, and also please rename the two migration scripts as outlined below. Thanks!

fghaas (approved):

fghaas:
This looks good to me now — do you want to give it another test-deploy to see that a newly created stack gets its `delete_by` attribute set by via the default hook, as intended?

mrtmm:
Yes, I'll do a test round in the morning.

PR #137, opened 2020-11-12
Fix support for nested video elements

fghaas:
edx-platform PR 20384 removed the VideoDescriptor class, and replaced it with VideoBlock.

PR reference: https://github.com/edx/edx-platform/pull/20384/

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,3 +1,8 @@
+Unreleased
+---------------------------
+
+* [Bug fix] Fix support for nested `<video>` elements
+
 Version 4.0.0 (2020-11-10)
 ---------------------------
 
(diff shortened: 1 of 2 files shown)


mrtmm (approved):

fghaas:
@mrtmm: I must confess I don't yet know how to test this. We evidently don't have a working test for this, otherwise we would have caught this earlier — if I compare the history of when we added this feature and when the corresponding change landed in edx-platform, this can't ever have actually worked on a contemporary Open edX platform.

I've tried adding a test but it would appear that the `xmodule` path can't be made to properly import in the test harness. However, I do know that the `VideoDescriptor` reference is _definitely_ broken, so this change doesn't make anything worse, even though it might not make anything better either.

PR #139, opened 2020-11-23
Fix hook_events import

mrtmm:
hook_events is a dictionary that contains boolean values. In course
export those values are converted to strings, but were not properly
parsed back to boolean values in course import.

Use distutils.util.strtobool to make sure that the boolean values in
hook_events are properly parsed in course import.

### DIFF
--- a/hastexo/hastexo.py
+++ b/hastexo/hastexo.py
@@ -14,6 +14,7 @@ from xblockutils.studio_editable import (
 )
 from xblockutils.settings import XBlockWithSettingsMixin
 
+from distutils.util import strtobool
 from django.db import transaction
 from django.utils import timezone
 from lxml import etree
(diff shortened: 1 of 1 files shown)


hastexo/hastexo.py
```diff
@@ -220,6 +220,11 @@ def parse_attributes(tag, node, block):
                         "environment": environment}
             block.providers.append(provider)
 
+    def to_bool(string_value):
```
fghaas, line 223: Any reason for rolling your own here, rather than using

```python
from distutils.util import strtobool

def to_bool(string_value):
  return bool(strtobool(value))
```

That approach would cover you for `true/TRUE/True/tRuE/yes/1` and `false/FALSE/False/FaLsE/0`.

hastexo/hastexo.py
```diff
@@ -252,11 +257,11 @@ def parse_xml(cls, node, runtime, keys, id_generator):
 
                     elif child.tag == "hook_events":
                         hook_events = {
-                            "suspend": bool(
+                            "suspend": cls.to_bool(
```
fghaas, line 260: ... or else, drop the wrapper method and just open-code `bool(strtobool(<value>))` here.

mrtmm: Updated the PR with this.

fghaas (approved):
Looks good to me now, thanks!

PR #141, opened 2020-12-11
Add an option to display a terminal in read-only mode

mrtmm:
This commit adds an optional XBlock attribute "read_only",
that allows to display a lab terminal in read-only mode.
If set to True, a stack will be created or resumed as usual,
the student can see the terminal but is unable to interact with it.
The value defaults to False.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,3 +1,8 @@
+Unreleased
+----------
+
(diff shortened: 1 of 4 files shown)


fghaas (approved):
Looks great, thanks!

PR #145, opened 2021-01-13
Only add attributes to a provider when they have a value

mrtmm:
Only add `template` and  `environment` attributes to a
provider during course export and import when they have a value.
Adding a `null` or `None` values can cause breakage with spinning up labs.

### DIFF
--- a/hastexo/hastexo.py
+++ b/hastexo/hastexo.py
@@ -212,19 +212,23 @@ class HastexoXBlock(XBlock,
             name = node.attrib["name"]
             if not name:
                 raise KeyError("name")
-            capacity = node.attrib.get("capacity")
+            capacity = node.attrib.get("capacity", None)
             if capacity in (None, "None", ""):
+                # capacity should not be undefined
+                # set to -1 (unlimited) in case it is
                 capacity = -1
             else:
                 # This will raise a TypeError if the string literal
                 # cannot be converted
                 capacity = int(capacity)
-            template = node.attrib.get("template", None)
-            environment = node.attrib.get("environment", None)
             provider = {"name": name,
-                        "capacity": capacity,
-                        "template": template,
-                        "environment": environment}
+                        "capacity": capacity}
+            template = node.attrib.get("template", None)
+            if template not in (None, "None"):
+                provider["template"] = template
+            environment = node.attrib.get("environment", None)
+            if environment not in (None, "None"):
+                provider["environment"] = environment
             block.providers.append(provider)
 
     @classmethod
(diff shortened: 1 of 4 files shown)


hastexo/hastexo.py
```diff
@@ -219,12 +219,13 @@ def parse_attributes(tag, node, block):
                 # This will raise a TypeError if the string literal
                 # cannot be converted
                 capacity = int(capacity)
-            template = node.attrib.get("template", None)
             environment = node.attrib.get("environment", None)
             provider = {"name": name,
                         "capacity": capacity,
-                        "template": template,
                         "environment": environment}
+            template = node.attrib.get("template", None)
```
fghaas, line 226: Let's please also apply the same logic to `environment` and `capacity`.

fghaas (changes requested):

hastexo/hastexo.py
mrtmm: Ack, sure.

hastexo/hastexo.py
```diff
@@ -212,19 +212,16 @@ def parse_attributes(tag, node, block):
             name = node.attrib["name"]
             if not name:
                 raise KeyError("name")
-            capacity = node.attrib.get("capacity")
-            if capacity in (None, "None", ""):
-                capacity = -1
```
fghaas, line 217: Hold on, I'm not sure if removing this is correct. The logic here is meant to be:

- if `capacity` is set to a negative value, then it's unlimited
- if it's set to 0, then it's actually 0 (i.e. don't run anything here)
- if it's set to a positive value, then that's the actual limit.

Are you sure your changes preserve that?

mrtmm: This change drops the `capacity` attribute when it is for some reason set to None, "None" or "". If it is set to a negative value or 0 these values will be exported and imported as they are. However, I am not sure if the absence of the `capacity` attribute altogether will make spinning up labs to fail. If so then we might not be able to apply the same logic for `capacity` as we do for `template`.

fghaas:
Would you mind rebasing this topic branch on `master` as well, so that we can trust the status checks on this PR?

hastexo/hastexo.py
fghaas: OK, it would be good to have a regression test for this. Can you please update `test_tasks.py`, adding test cases for an undefined `capacity`?

fghaas (changes requested):

hastexo/hastexo.py
mrtmm: Ok, we cannot have `capacity` undefined, labs will fail with that. I reverted this part and added additional tests for it.

fghaas (approved):

fghaas:
This looks good to me now, thanks!

PR #150, opened 2021-01-29
Add more configuration options for progress check result dialog

mrtmm:
Add a new XBlock attribute `progress_check_result_heading` for
configuring the progress check result dialog message.
Add a new XBlock attribute `show_feedback` to enable/disable
showing feeback for tasks completion.

### DIFF
--- a/README.md
+++ b/README.md
@@ -521,8 +521,17 @@ The following are optional:
 * `progress_check_label`: Set a label for the progress check button.
   For example: `Submit Answer` or `Check Progress` (Default).
 
+* `show_feedback`: On progress check, show feedback on how many tasks out of total 
+  are completed. Default is `True`.
+
 * `show_hints_on_error`: On progress check failure, display the tests' standard
-  error streams as hints. Default is `True`.
+  error streams as hints.
+  When `show_feedback` is set to `False`, hints will never be displayed and having
+  this set to `True` will have no effect. Default is `True`.
+
+* `progress_check_result_heading`: Message to display on progress check result window.
+  This could be set to "Answer Submitted" for example, when choosing to not display
+  hints and feedback. Default is "Progress check result".
 
 You can also use the following nested XML options:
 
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -524,6 +524,15 @@ The following are optional:
 * `show_hints_on_error`: On progress check failure, display the tests' standard
   error streams as hints. Default is `True`.
 
+* `show_feedback`: On progress check, show feedback on how many tasks out of total 
+  are completed. When `show_hints_on_error` is set to `True`, feedback will always
+  be displayed and setting it to `False` here will not take effect.
+  Default is `True`.
```
fghaas, line 530: Intuitively I think of `show_feedback` as being higher-order than `show_hints_on_error`, so my thought would be to flip the logic (and the order) here and instead make it such:

```markdown
* `show_feedback`: On progress check, show feedback on how many tasks out of total 
  are completed. 

* `show_hints_on_error`: On progress check failure, display the tests' standard
  error streams as hints.
  When `show_feedback` is set to `False`, hints will never be displayed and setting
  this to `True` will have no effect.
```

Also I think you don't need to put "Default is" into the prose as Studio will show the default anyway.

fghaas (changes requested):
Thanks! I have one comment. Please consider that one and either implement the change, or convince me that your approach is better, UX wise. :)

README.md
mrtmm: I think that is a good point, I will implement the change.

hastexo/hastexo.py
```diff
@@ -110,11 +110,27 @@ class HastexoXBlock(XBlock,
         help="Set the progress check button label. "
              "For example: \"Submit Answer\" or \"Check Progress\"(Default)."
     )
+    show_feedback = Boolean(
+        default=True,
+        scope=Scope.settings,
+        help="On progress check, show feedback on how many tasks out of total "
+             "are completed. "
```
fghaas, line 117: Micro-nit: is this trailing whitespace after "completed." intentional?

fghaas (commented):
Looks good to me now except for a minute typo.

hastexo/hastexo.py
mrtmm: Not intentional, deleted now.

PR #152, opened 2021-02-01
Fix exporting provider attributes 

mrtmm:
Provider attributes like `template` and `environment` will not be
added to the provider node during course export in case of None or
"None" values.
However, not defining those attributes at all will throw an error
during export.
Add an additional check to make sure export succeeds when these attributes
are not defined for a provider.

### DIFF
--- a/hastexo/hastexo.py
+++ b/hastexo/hastexo.py
@@ -375,12 +375,19 @@ class HastexoXBlock(XBlock,
             if self.providers:
                 for provider in self.providers:
                     provider_node = etree.SubElement(root, 'provider')
-                    provider_template = provider.get("template", None)
-                    if provider_template in (None, "None"):
-                        provider.pop("template")
-                    provider_environment = provider.get("environment", None)
-                    if provider_environment in (None, "None"):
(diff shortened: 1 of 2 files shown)


hastexo/hastexo.py
```diff
@@ -375,12 +375,15 @@ def add_xml_to_node(self, node):
             if self.providers:
                 for provider in self.providers:
                     provider_node = etree.SubElement(root, 'provider')
-                    provider_template = provider.get("template", None)
-                    if provider_template in (None, "None"):
-                        provider.pop("template")
-                    provider_environment = provider.get("environment", None)
-                    if provider_environment in (None, "None"):
-                        provider.pop("environment")
+                    if 'template' in provider:
```
fghaas, line 378: Maybe you want to add some of the information as to _why_ you're doing it this way, as a comment? That would be a nice touch for Future Maari who comes back at this in a year or two, and scratches her head why the extra check is there. :)

mrtmm: Sounds good, added a comment.

PR #155, opened 2021-02-10
Make the XBlock a subclass of ScorableXBlockMixin

mrtmm:


fghaas (commented):
Looks good to me! Just making sure: no migrations with this one?

mrtmm:
No migrations but might need a memcached restart for correct grading to kick in.

PR #156, opened 2021-02-11
Fix grading related instructor tasks

mrtmm:
Implement additional `ScorableXBlockMixin` methods for using the
grading related instructor tasks like overriding and rescoring
learner's submissions.

### DIFF
--- a/hastexo/hastexo.py
+++ b/hastexo/hastexo.py
@@ -924,8 +924,9 @@ class HastexoXBlock(XBlock,
                     status = result.result
                     score = Score(raw_earned=status['pass'],
                                   raw_possible=status['total'])
+                    self.set_score(score)
                     # A publish event is necessary for calculating grades
-                    self.publish_grade(score)
+                    self.publish_grade()
 
                 else:
                     status = {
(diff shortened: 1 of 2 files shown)


hastexo/hastexo.py
```diff
@@ -992,10 +999,37 @@ def set_score(self, score):
...
-    def publish_grade(self, score):
+    def get_score(self):
+        """
+        Return a raw score already persisted on the XBlock.  Should not
+        perform new calculations.
+        Returns:
+            Score(raw_earned=float, raw_possible=float)
+        """
+        if not self.score:
+            logger.warning("No score is earned for this block yet")
```
fghaas, line 1010: Just checking: is this really a "warning" condition? Wouldn't we call this every time an instructor downloads a grade report, rather needlessly flooding the logs? (Consider what happens in a course with 1,000 enrolled learners.)

fghaas (commented):
Looks good to me, I have one tiny nit.

hastexo/hastexo.py
mrtmm: Good question, I am actually not sure how to download a grade report. I believe the grades are generally fetched from the database so as far as I know we are only calling this after a score is set from a student's submission or from instructor's override, so we really shouldn't see this warning at all, but I added it for future us to help tracing down some unexpected behaviour. I really don't know at this point if we should be returning None or maybe a default 0/max_score here.

PR #157, opened 2021-02-16
Allow overriding hastexo_guacamole_client settings

mrtmm:
Allow overriding hastexo_guacamole_client settings.py values from
a configuration file by defining the file as `HASTEXO_GUACAMOLE_CFG`

### DIFF
--- a/hastexo_guacamole_client/settings.py
+++ b/hastexo_guacamole_client/settings.py
@@ -1,24 +1,19 @@
 import os
+from distutils.util import strtobool
+from yaml import load, SafeLoader
+from yaml.scanner import ScannerError
 
 from django.core.exceptions import ImproperlyConfigured
-
-
-def get_env_setting(setting):
-    """ Get the environment setting or return exception """
-    try:
-        return os.environ[setting]
-    except KeyError:
-        error_msg = u"Set the %s env variable" % setting
-        raise ImproperlyConfigured(error_msg)
 
 
 BASE_DIR = os.path.dirname(os.path.dirname(os.path.dirname(__file__)))
 
 # SECURITY WARNING: keep the secret key used in production secret!
-SECRET_KEY = get_env_setting('HASTEXO_GUACAMOLE_SECRET_KEY')
+SECRET_KEY = os.environ.get('HASTEXO_GUACAMOLE_SECRET_KEY')
 
 # SECURITY WARNING: don't run with debug turned on in production!
-DEBUG = os.getenv('HASTEXO_GUACAMOLE_DEBUG', False)
+DEBUG = bool(strtobool(os.environ.get('HASTEXO_GUACAMOLE_DEBUG', 'false')))
+DJANGO_LOG_LEVEL = os.environ.get('HASTEXO_GUACAMOLE_LOG_LEVEL', 'WARNING')
 
 LOGGING = {
     'version': 1,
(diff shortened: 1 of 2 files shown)


hastexo_guacamole_client/settings.py
```diff
@@ -36,12 +31,28 @@ def get_env_setting(setting):
...
+DATABASES = {
+    'default': {
+        'ENGINE': 'django.db.backends.mysql',
+        'NAME': os.environ.get('HASTEXO_GUACAMOLE_DEFAULT_DB_NAME', 'edxapp'),
+        'USER': os.environ.get('HASTEXO_GUACAMOLE_DATABASE_USER', 'edxapp001'),
+        'PASSWORD': os.environ.get(
+            'HASTEXO_GUACAMOLE_DATABASE_PASSWORD', 'password'),
+        'HOST': os.environ.get('HASTEXO_GUACAMOLE_DATABASE_HOST', 'localhost'),
+        'PORT': int(os.environ.get('HASTEXO_GUACAMOLE_DATABASE_PORT', '3306')),
+        'ATOMIC_REQUESTS': os.environ.get(
```
fghaas, line 49: Here too I think you need `bool(strtobool())`, otherwise the string `false` will be considered a truthy value.

hastexo_guacamole_client/settings.py
```diff
@@ -1,24 +1,19 @@
...
-        error_msg = u"Set the %s env variable" % setting
-        raise ImproperlyConfigured(error_msg)
-
-
 BASE_DIR = os.path.dirname(os.path.dirname(os.path.dirname(__file__)))
 
 # SECURITY WARNING: keep the secret key used in production secret!
-SECRET_KEY = get_env_setting('HASTEXO_GUACAMOLE_SECRET_KEY')
+SECRET_KEY = os.environ.get('HASTEXO_GUACAMOLE_SECRET_KEY',
+                            'insecure-secret-key')
```
fghaas, line 12: Not a big fan of setting a default here, to be honest. If you remove the default, Django will blow up on startup with `ImproperlyConfigured` if `SECRET_KEY` is unset, which I'd argue is the behavior you want.

hastexo_guacamole_client/settings.py
```diff
@@ -1,24 +1,19 @@
...
 
 # SECURITY WARNING: don't run with debug turned on in production!
-DEBUG = os.getenv('HASTEXO_GUACAMOLE_DEBUG', False)
+DEBUG = os.environ.get('HASTEXO_GUACAMOLE_DEBUG', False)
```
fghaas, line 15: I believe you need to wrap this in `bool(strtobool())`, otherwise setting the environment variable to the string `false` will result in `DEBUG = True`.

fghaas (changes requested):
Looks good overall; I see a couple of issues with envar parsing though.

hastexo_guacamole_client/settings.py
mrtmm: Ack, done.

hastexo_guacamole_client/settings.py
mrtmm: done.

hastexo_guacamole_client/settings.py
mrtmm: done.

hastexo_guacamole_client/settings.py
```diff
@@ -1,24 +1,19 @@
...
-
 BASE_DIR = os.path.dirname(os.path.dirname(os.path.dirname(__file__)))
 
 # SECURITY WARNING: keep the secret key used in production secret!
-SECRET_KEY = get_env_setting('HASTEXO_GUACAMOLE_SECRET_KEY')
+SECRET_KEY = os.environ.get('HASTEXO_GUACAMOLE_SECRET_KEY')
 
 # SECURITY WARNING: don't run with debug turned on in production!
-DEBUG = os.getenv('HASTEXO_GUACAMOLE_DEBUG', False)
+DEBUG = bool(strtobool(os.environ.get('HASTEXO_GUACAMOLE_DEBUG', False)))
```
fghaas, line 15: This won't work for an unset envar:

```doctest
>>> bool(strtobool(os.environ.get('HASTEXO_GUACAMOLE_DEBUG', False)))
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3.8/distutils/util.py", line 313, in strtobool
    val = val.lower()
AttributeError: 'bool' object has no attribute 'lower'
```
The default value passed into `os.environ.get()` has to be `'false'` (string), not `False` (boolean).

fghaas (changes requested):

hastexo_guacamole_client/settings.py
mrtmm: Done.

PR #158, opened 2021-02-18
Updates to `hastexo_guacamole_client`

mrtmm:
- Add `hastexo_guacamole_client` requirements
- Refactor `asyncio` task creation logic in `consumers.py` to make things also work with python3.5
- Provide the `stack_protocol` attribute when initializing javascript to _unbreak_ calculating the terminal height correctly
- Implement the `read_only` mode in the websocket consumer instead of having a (possibly editable) switch in javascript code


requirements/base.txt
```diff
@@ -12,6 +12,11 @@ python-heatclient>=1.6.1,<1.7
 python-keystoneclient>=3.17.0,<3.18
 python-novaclient>=7.1.2,<7.2
 tenacity>=6.2
+
+# for hastexo_guacamole_client
 channels>=2.4.0,<3
-pyguacamole>=0.9
+asgiref==3.2.0  # depends on our channels version
+mysqlclient==1.3.14  # keep in sync with edx-platform
+jsonfield2==3.0.3   # keep in sync with edx-platform
```
fghaas, line 20: Any chance we can be a little more lenient here, as in using a sensible combination of `>=` and `<`, so as to not break things if edx decides to up the patch level of one of these?

mrtmm: Sure, I've gone through the release histories and made some assumptions in 589344 for what should not break things for us _now_ but I'll just have try it out.

Changelog.md
```diff
@@ -1,5 +1,7 @@
 Unreleased
 -----------------------------
+* [Bug fix] Refactor asynio task creation logic to also work with
```
fghaas, line 9: Micro-nit: typo. :)

hastexo_guacamole_client/consumers.py
```diff
@@ -69,7 +72,11 @@ async def receive(self, text_data=None, bytes_data=None):
         Handle data received in the WebSocket, send to GuacamoleClient.
         """
         if text_data is not None:
-            self.client.send(text_data)
+            # ignore all 'key' events when set to 'read_only" mode
```
fghaas, line 75: Just checking: do `key` events include mouse clicks? If not, do we want to disable those as well?

fghaas (commented):
Just a thought on this bit:

> How we calculate the terminal height depends on the `stack_protocol` attribute.

We might want to rethink that at some point. Meaning, I'm no longer sure if the default geometry should differ, based on whether we have a terminal (`ssh`) or desktop (`rdp`). But for now, let's preserve the prior behavior.

I have one tiny nit, and a question. Otherwise this looks great to me.

Changelog.md
mrtmm: Thanks, fixed.

hastexo_guacamole_client/consumers.py
mrtmm: Disabled mouse events as well now.

PR #160, opened 2021-02-23
Fix `Stack.DoesNotExist` error when deleting learner state

mrtmm:
When deleting learner state, the `stack_name` value
gets wiped out. If we then try to update the stack (for example
via `keepalive`) we get `Stack.DoesNotExist` error. Check if
`stack_name` has a value before attempting to update and if not,
set it again.

### DIFF
--- a/hastexo/hastexo.py
+++ b/hastexo/hastexo.py
@@ -514,6 +514,12 @@ class HastexoXBlock(XBlock,
             # delete_age value in settings is in days, convert to seconds
             return settings.get("delete_age", 14) * 86400
 
+    def get_stack_name(self):
+        # Get the course id and anonymous user id, and derive the stack name
+        # from them
+        course_id, student_id = self.get_block_ids()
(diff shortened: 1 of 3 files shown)


hastexo/hastexo.py
```diff
@@ -678,6 +684,10 @@ def update_stack(self, data):
         in a transaction.
         """
         course_id, student_id = self.get_block_ids()
+        if not self.stack_name:
+            # Stack name can be occasionally end up being empty (after deleting
+            # learners state for example). If so, set it again.
+            self.stack_name = self.get_stack_name()
```
fghaas, line 690: Looks good. Can you think of a way to fix up a unit test so we get some coverage on this `if` block?

mrtmm: Sure, I'll check that now.

mrtmm: Added a test.

fghaas (approved):
Looks good to me now, thanks!

PR #163, opened 2021-03-17
Retry on Paramiko socket timeout

fghaas:
When we don't get an SSH connection from Paramiko immediately, we don't want to wait forever until we hit the task timeout. Instead, we define a shorter timeout, `ssh_connect_timeout`, defaulting to 10 seconds, and use that to set a socket timeout option with the Paramiko [`connect()`](http://docs.paramiko.org/en/stable/api/client.html#paramiko.client.SSHClient.connect) call.

If we fail to establish a connection within that time, we hit the [`socket.timeout`](https://docs.python.org/3/library/socket.html#socket.timeout) error, which we then catch in order to retry the connection.

### DIFF
--- a/.github/workflows/tox.yml
+++ b/.github/workflows/tox.yml
@@ -9,7 +9,9 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.6, 3.7, 3.8]
+        python-version:
+          - 3.5
+          - 3.8
 
     steps:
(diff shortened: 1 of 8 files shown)


mrtmm (approved):
Looks good to me.

fghaas:
@mrtmm Are you still good with this?

mrtmm:
Yes, looks good to me!

PR #167, opened 2021-03-25
Updates for Open EdX Koa release

mrtmm:


fghaas (approved):

fghaas:
Looks good to me, feel free to merge at your convenience.

hastexo/admin.py
```diff
@@ -3,7 +3,7 @@
 """
 from django import forms
 from django.contrib import admin
-from student.models import AnonymousUserId
+from common.djangoapps.student.models import AnonymousUserId
```
fghaas, line 6: So, just checking: do we know that this import also works on edx-platform Juniper? (I ask because the tests can't tell us, since we deliberately fake this out).

If it does not, I suggest we change this to:

```python
try:
    from common.djangoapps.student.models import AnonymousUserId
except ImportError:
    from student.models import AnonymousUserId

File: tests/unit/test_admin.py [hunk 2]
```diff
@@ -8,7 +8,7 @@
 from django.test import TestCase
 from django.test.client import Client
 from django.urls import reverse
-from student.models import AnonymousUserId
+from common.djangoapps.student.models import AnonymousUserId
```
fghaas, line 11: Same applies here.

hastexo/admin.py
```diff
@@ -3,7 +3,10 @@
 """
 from django import forms
 from django.contrib import admin
-from student.models import AnonymousUserId
+try:
+    from common.djangoapps.student.models import AnonymousUserId
+except RuntimeError:
```
fghaas, line 8: I'm confused. Why exactly do we need `RuntimeError` here, instead of `ImportError`?

mrtmm: I tested this in devstack and I got a RuntimeError instead of ImportError.

PR #171, opened 2021-04-15
Make the reset warning more verbose

fghaas:
* Add a reminder that resetting a lab will take several minutes.
* Add an additional sentence (set to `display: none;`) reminding a learner that if they are in a middle of an exam, the exam timer will   continue to run while they are resetting.

What's still left here is some JavaScript magic that flips the `display` property when there is an exam timer (`<span class="exam-timer-clock">`) somewhere on the page.

### DIFF
--- a/.github/workflows/tox.yml
+++ b/.github/workflows/tox.yml
@@ -10,7 +10,6 @@ jobs:
     strategy:
       matrix:
         python-version:
-          - 3.5
           - 3.8
 
     steps:
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,5 +1,7 @@
 Unreleased
 -------------------------
+* [Enhancement] Display the "check progress" button (if enabled) in
+  blue, to provide greater contrast to the red reset button.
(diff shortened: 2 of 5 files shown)


fghaas:
No idea why the workflow run is failing. I suspect a problem with GitHub's pip cache.

mrtmm (approved):

fghaas:
@mrtmm I've come to the tentative conclusion that GitHub Actions are evidently broken for setuptools_scm on Python 3.5, and that the package name we use (`setuptools-scm` vs. `setuptools_scm`) is irrelevant. I've thus disabled the Python 3.5 pipeline in GitHub actions — though it remains in `tox.ini`, since it works perfectly fine on my local development setup. See cccd043.

If this looks good to you, I'd now go ahead and merge this and cut a fresh point release.

mrtmm:
I think we are missing a changelog entry, other than that, this looks good to me!

fghaas:
Yes, absolutely. Two, actually. :) I've added those now.

PR #172, opened 2021-04-16
Register Celery tasks to an app

mrtmm:
Starting with Celery version 4.0 the Task base class no longer
automatically register tasks (https://docs.celeryproject.org/en/4.0/whatsnew-4.0.html#the-task-base-class-no-longer-automatically-register-tasks)
and we will need to register our class based tasks "manually".


fghaas (changes requested):
This looks good to me in principle, but the test failures should be fixed before we can merge this.

PR #177, opened 2021-04-22
Make an extra warning message visible for timed exams

mrtmm:


fghaas:
Thanks! I'll squash that into #171 now.

PR #178, opened 2021-04-23
Update requirements

mrtmm:


requirements/base.txt
```diff
@@ -2,15 +2,15 @@ apscheduler>=3.5.1,<3.6
 cliff>=2.2.0,<2.3
```
fghaas, line 2: Are we sure the explicit requirement for `cliff` is still needed here?

requirements/base.txt
```diff
@@ -1,17 +1,16 @@
...
-python-heatclient>=1.6.1,<1.7
-python-keystoneclient>=3.17.0,<3.18
-python-novaclient>=7.1.2,<7.2
-tenacity>=6.2
+keystoneauth1>=3.13.0,<3.18
+osc-lib>=1.2.0,<2
+os-client-config>=1.28.0,<2
+oslo.serialization>=2.28.1,<3
+oslo.utils>=3.37.1,<3.42.1
+paramiko>=2.1.6,<=2.7.2
```
fghaas, line 9: Don't use `<=<point release>` here; you'd miss critical patch updates within the minor release series. Use `<2.8` instead.

requirements/base.txt
```diff
@@ -1,17 +1,16 @@
...
+python-heatclient>=1.6.1,<2
+python-keystoneclient>=3.17.0,<3.22
+python-novaclient>=7.1.2,<16
+tenacity>=6.2,<=7.0.0
```
fghaas, line 13: Here too. Use `<7.1`, or even `<8`.

requirements/base.txt
```diff
@@ -1,17 +1,16 @@
...
-paramiko>=2.1.6,<2.2
-python-heatclient>=1.6.1,<1.7
-python-keystoneclient>=3.17.0,<3.18
-python-novaclient>=7.1.2,<7.2
-tenacity>=6.2
+keystoneauth1>=3.13.0,<3.18
+osc-lib>=1.2.0,<2
+os-client-config>=1.28.0,<2
+oslo.serialization>=2.28.1,<3
+oslo.utils>=3.37.1,<3.42.1
```
fghaas, line 8: Any particular reason why you'd allow 3.42.0 here, but not 3.42.1? Per semver the only difference between .0 and .1 could possibly be that .1 contains a patch fixing a bug that .0 has.

fghaas (changes requested):
Looks good except for 3 tiny nits.

requirements/base.txt
mrtmm: Should this be <=3.42.1 then or <3.43?

fghaas: Per `oslo.utils`' [release history](https://pypi.org/project/oslo.utils/#history), if OpenStack Train is our reference point, I'd suggest `<4`.

requirements/base.txt
mrtmm: Done.

fghaas (approved):
Looks good to me now, thanks!

requirements/base.txt
mrtmm: Done.

requirements/base.txt
mrtmm: Ack, done.

PR #180, opened 2021-05-12
Speed up progress checks

mrtmm:
Reduce the sleep time to 1s when waiting for a remote execution
of a command to finish.

Add an except clause for RemoteExecTimeout so that the timeout
message would be displayed as "Unexpected result" instead of
as a hint to the learner.


fghaas (approved):
Looks good to me!

PR #182, opened 2021-05-20
Fix RDP connectivity checks for IPv6 stacks

mrtmm:
Our RDP connectivity check is using `socket.socket().connect()`,
which will only try an IPv4 connection and will break for stacks
exposing an IPv6 `public_ip`.
Use `socket.create_connection()` instead, which will try both,
IPv4 and IPv6 connection.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,3 +1,7 @@
+Unreleased
+-------------------------
+* [Bug fix] Fix RDP connectivity check for IPv6 stacks.
+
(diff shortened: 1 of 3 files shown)


fghaas:
Change looks great; commit message subject needs an update. :) (It's not RDP connectivity in general that's broken, just the connectivity _check,_ and even that only for IPv6.)

fghaas (approved):
Looks great now, thanks!

PR #183, opened 2021-05-20
Sanitize stack names

mrtmm:
We build the `stack_name` based on `course_id` and `student_id`.
We also use the same `stack_name` for creating a keypair for the
learner. However, Nova has a list of "unsafe" characters which
will throw an exception, if used in the keypair name.
(https://github.com/openstack/nova/blob/master/nova/compute/api.py#L6396)

Make sure that the `stack_name` consists of only safe characters.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,3 +1,7 @@
+Unreleased
+-------------------------
+* [Bug fix] Restrict stack names to ASCII characters and digits.
+
 Version 5.0.13 (2021-05-21)
 -------------------------
 * [Bug fix] Fix RDP connectivity check for IPv6 stacks.
--- a/hastexo/hastexo.py
+++ b/hastexo/hastexo.py
@@ -1,6 +1,8 @@
 import time
 import logging
 import os
+import re
+import string
 import textwrap
 
 from xblock.core import XBlock, XML_NAMESPACES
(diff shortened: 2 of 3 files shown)


hastexo/hastexo.py
```diff
@@ -518,7 +519,14 @@ def get_stack_name(self):
         # Get the course id and anonymous user id, and derive the stack name
         # from them
         course_id, student_id = self.get_block_ids()
-        return "%s_%s_%s" % (course_id.course, course_id.run, student_id)
+        stack_name = "%s_%s_%s" % (course_id.course, course_id.run, student_id)
+
+        # make sure that the stack name doesn't include any unsafe characters
+        # (according to Nova, for creating a keypair)
+        safe_chars = "_- " + string.digits + string.ascii_letters
+        stack_name = "".join(x if x in safe_chars else '_' for x in stack_name)
```
fghaas, line 527: While Nova may allow whitespace and hyphens in keypair names, I'm not entirely sure if we want them in stack names.

I'd just go with this:

```python
# Replace anything in the stack name that is not an ASCII letter or digit with an underscore
replace_pattern = '[^%s%s]' % (string.digits, string.ascii_letters)
stack_name = re.sub(re.compile(replace_pattern), '_', stack_name)
```

What do you think?

mrtmm: Sounds good,  I had my doubts about the whitespace as well.

Changelog.md
```diff
@@ -1,3 +1,8 @@
+Unreleased
+-------------------------
+* [Bug fix] make sure that the `stack_name` doesn't include any
```
fghaas, line 3: For the changelog, I'd just put "Restrict stack names to ASCII characters and digits." The detail about Nova and keypairs is already in the commit message, where it belongs.

mrtmm: Ack, done.

PR #184, opened 2021-05-28
Add tests to cover import/export logic; Make XBlock exports deterministic and predictable.

mrtmm:
* Add tests for new export/import logic
* Make XBlock exports deterministic and predictable.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,5 +1,6 @@
 Unreleased
 -------------------------
+* [Enhancement] Add tests for the new export/import logic.
 * [Bug fix] Restrict stack names to ASCII characters and digits.
 
 Version 5.0.13 (2021-05-21)
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,5 +1,7 @@
 Unreleased
 -------------------------
+* [Bug fix] Make XBlock exports (from Studio or its REST API)
+  deterministic and predictable.
 * [Enhancement] Add tests for the new export/import logic.
 * [Bug fix] Restrict stack names to ASCII characters and digits.
 
--- a/hastexo/hastexo.py
+++ b/hastexo/hastexo.py
@@ -237,21 +237,16 @@ class HastexoXBlock(XBlock,
             block.tests.append(text)
 
         elif tag == "port":
-            name = node.attrib["name"]
-            if not name:
-                raise KeyError("name")
-            number = node.attrib["number"]
-            if not number:
-                raise KeyError("number")
-            number = int(number)
-            port = {"name": name,
-                    "number": number}
+            # port must have values for 'name' and 'number'
+            # raises KeyError for each if not defined
+            port = {"name": node.attrib["name"],
+                    "number": int(node.attrib["number"])}
             block.ports.append(port)
 
         elif tag == "provider":
+            # raises KeyError if 'name' is not defined
+            # one must not add a provider without a name
             name = node.attrib["name"]
-            if not name:
-                raise KeyError("name")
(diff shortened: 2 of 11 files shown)


Changelog.md
```diff
@@ -1,5 +1,7 @@
 Unreleased
 -------------------------
+* [Bug fix] Remove randomness on how the XBlock attributes are
```
fghaas, line 3: "Remove randomness" sounds a little awkward. :) I'd put "Make XBlock exports (from Studio or its REST API) deterministic and predictable."

hastexo/hastexo.py
```diff
@@ -343,14 +340,6 @@ def parse_xml(cls, node, runtime, keys, id_generator):
 
         return block
 
-    def add_dict_properties_to_node(self, properties, node):
-        """
-        Add properties from a dictionary as attributes to `node`.
-
-        """
-        for key in properties.keys():
-            node.set(key, str(properties.get(key, '')))
```
fghaas, line 352: Just a question: I guess another way to do this would be to just keep using this method, albeit with

```python
for key in sorted(properties.keys()):
```

... which would sort the keys asciibetically. But I'm sure you thought of that, so I'm curious what advantages you found out the explicit open-coded approach has. :)

fghaas (commented):
This is excellent! If you could throw in a `fake_lab_3.xml` with a nested `<markdown>` element we'd even get test coverage for import/export of nested blocks. :)

hastexo/hastexo.py
mrtmm: Mainly to handle a missing key with either throwing a KeyError (port name & number, provider name) or setting a default (True for hook_events & -1 for capacity). It ended up looking more readable like this for me but I can give it another look.

mrtmm:
So the nested xblocks go to `/vertical` right now not to the separate file. We have some tests for that but I can add more. Would we want the nested xblocks to go to the separate file?

Changelog.md
mrtmm: Ack, done.

hastexo/hastexo.py
fghaas: No, that's fine. More readable is good. :)

fghaas:
> So the nested xblocks go to `/vertical` right now not to the separate file. We have some tests for that but I can add more.

OK so if you look at a `coverage` report, you'll see that (in `hastexo.py`), the following bit appears to have no test coverage:

```python
        # Include nested blocks in course export 
        if self.has_children: 
            for child_id in self.children: 
                child = self.runtime.get_block(child_id) 
                self.runtime.add_block_as_child_node(child, node) 
```

So that's why I figured we might be missing that bit in the export test. Am I mistaken?

mrtmm:
Ah, I see, we only cover importing nested xblocks now. I'll take a look at export.

Changelog.md
```diff
@@ -1,5 +1,6 @@
 Unreleased
 -------------------------
+* [Enhacement] Add tests for the new export/import logic.
```
fghaas, line 5: Micro-nit: typo, missing an "n" here.

mrtmm:
I've added an export test with a nested xblock now.

fghaas (commented):
Splendid! I have two micro-nits: you're missing an "n" in the changelog, and the full stop at the end of the ba03dde commit message needs to go. Otherwise, this looks great!

Changelog.md
mrtmm: Thanks, done.

PR #186, opened 2021-06-07
WIP: Upload coverage data to Codecov

fghaas:
Add the ability to upload coverage data to Codecov, from the
GitHub Actions workflow.

PR #187, opened 2021-06-11
Truncate the error message for LaunchError

mrtmm:
When a stack launch is unsuccessful, we want to add the error message to
the Stack object's `error_msg` field, which has a limit of 256 characters.
Truncate the `LaunchError` error message to fit that limit.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,3 +1,9 @@
+Unreleased
+-------------------------
+* [Bug fix] Truncate the error message for `LaunchError` to fit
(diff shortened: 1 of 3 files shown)


fghaas:
This looks excellent to me, and also it appears that the codecov issues we saw in #186 have been fixed (never mind the "unsuccessful" codecov check; that's just due to the fact that we haven't run codecov in ages.

I'm good to merge this, unless you have any second thoughts about it?

mrtmm:
I'm good with merging it.

PR #188, opened 2021-06-11
Final round of Stable 4.1 backports

fghaas:
Now that Lilac has been released, it sound like a good time to retire stable-4.1 and support only Koa and Lilac going forward.

Thus, backport the last few important fixes from master, so we can cut a final release and put this branch to rest.

### DIFF
--- a/.github/workflows/tox.yml
+++ b/.github/workflows/tox.yml
@@ -22,8 +22,10 @@ jobs:
         python-version: ${{ matrix.python-version }}
     - name: Install dependencies
       run: |
-        pip install tox tox-gh-actions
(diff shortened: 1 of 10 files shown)


mrtmm (approved):
Looks good to me, not sure about the test failure.

fghaas:
Yeah the test failures are basically a result of Python 3.5 being well beyond EOL. The last couple of commits fix the test; they're technically not backports because they are not in `master`, but we probably want to include them anyway.

mrtmm:
Ack, thanks!

fghaas:
4.1.15 is released to PyPI.

PR #189, opened 2021-08-30
Truncate error messages for stack update

mrtmm:
Make sure that any time we update the value of the `error_msg`
field of a `Stack`, we truncate the message we add before the
stack update.


fghaas (approved):
Looks good to me, thank you!

PR #193, opened 2022-01-04
Add constraints to Django version requirement

mrtmm:
The Hastexo XBlock depends on the Django version defined for `edx-platform`
but the `hastexo_guacamole_client` does not.
Add constraints to make sure we don't install a version we have not tested yet.


fghaas (approved):
Looks good to me, thank you! Do you want me to cut a fresh release straight away?

mrtmm:
I think so, yes :)

PR #194, opened 2022-01-04
Fix mismatched HTML tags

mrgnr:
A `<p>` was matched with a `</div>`, so fix by changing the `<p>` to `<div>`.


fghaas:
Thanks @mrgnr, good catch! This has highlighted a testing coverage gap which is being addressed in #195. When that merges, this PR will also be marked as merged.

PR #195, opened 2022-01-05
Add basic validation for static HTML content, and fix static HTML so it passes validation

fghaas:
This is an extension of PR #194, which it builds upon.

PR #194 showed that we do not have test coverage for the HTML template in `static`, allowing non-well-formed HTML to slip through and be packaged with the XBlock. Thus, add a unit test case that simply loads the file and parses it with `lxml.etree.HTMLParser` (with `recover=False` so the parser throws an exception on running into an HTML error).

Implementing that unit test then turned up an additional issue on top of the one that #194 fixes, namely incorrectly used `<input>` tags. Fix those as well.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,3 +1,9 @@
+Unreleased
+-------------------------
+* [Bug fix] Fix unbalanced tags (`<p>` vs. `<div>`) in the static
+  `main.html` template.
+* [Testing] Add basic HTML validation for static templates.
+
 Version 5.0.17 (2022-01-04)
 -------------------------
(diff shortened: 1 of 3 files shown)


mrtmm (approved):
Looks good to me, thanks!

fghaas:
Sorry, forgot to update the changelog. Done now.

fghaas:
In the course of adding the changelog, I found a small typo in a prior changelog entry and I figured that that didn't merit a separate PR so I am instead including it here. :)

mrtmm:
Good catch, thanks! :)

PR #196, opened 2022-02-10
Add support for multiple Apache Guacamole versions

mrtmm:
* Make the `guacamole-common-js` library version configurable by an added setting.
* Document the steps for adding support for new versions.
* Add support for Apache Guacamole version `1.4.0` and make it the new default.


mrtmm:
No, nothing broke. I just saw that we never updated the version and figured we might want to use matching versions of the JavaScript library and `guacd`. Also I just wanted to have it documented in a way, how we incorporate the the library and what version we are on. Also, we might want to incorporate some new features at some point (the Clipboard API comes to mind), for which we would need to use the new versions.

fghaas (changes requested):
The changes here are fine but please fix your spelling of "guacamole-js" in the commit message subject. :)

fghaas (approved):
Looks good to me now, thanks!

PR #197, opened 2022-02-10
Add support for Tutor deployment

mrtmm:
When deploying the edx-platform with Tutor, the functionality
for the `wait_for_ping` method breaks as the plugin is then
running inside a Docker container and unable to ping the newly
spun up lab.

We really don't need to rely on this logic as we are more interested
in the abilty to create an ssh or rdp connection to the lab.
We can drop the logic for it altogether.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,5 +1,7 @@
 Unreleased
 -------------------------
+* [Enhancement] Add support for Tutor deployment, by dropping the
+  `wait_for_ping` logic.
 * [Enhancement] Add support for Apache Guacamole version `1.4.0`,
   make it the new default.
 * [Enhancement] Make the `guacamole-common-js` library version
--- a/hastexo/tasks.py
+++ b/hastexo/tasks.py
@@ -1,9 +1,7 @@
 import time
-import os
 import traceback
 import socket
 import logging
-import ipaddress
 import textwrap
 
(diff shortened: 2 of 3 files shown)


Changelog.md
```diff
@@ -1,5 +1,7 @@
 Unreleased
 -------------------------
+* [Enhancement] Add support for Tutor deployment, by droping the
```
fghaas, line 3: Nit: typo, should be "dropping".

mrtmm: Thanks, done.

fghaas (approved):
Looks good to me now, thanks! Do you want to give this another round of testing, or do you want me to cut a release? I'd call this one 6.0.0rc1 so we can tell people "if you want to use this with Tutor, use version 6".

mrtmm:
I think we could merge this so we can test the plugin by installing this from master, not from my fork. I'd hold on with the release a bit longer, I have a few commits in mind to add before (dependency updates, a typo or two), if that's ok.

fghaas:
Yes, sure thing. This one's merged. :)

fghaas:
By the way: we can have as many `rc` tags as we would like; for the 5.0.0 release we had 6. :)

mrtmm:
Right, yeah, I guess in that case it doesn't matter when we cut the first one :)

PR #198, opened 2022-02-11
Update Requirements

mrtmm:
Update base and test requirements to keep in line with the
edx-platform.

Drop constraints for Django for test requirements as we
recently added the same in requirements/base.txt.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,6 +30,17 @@ SSH).
 > your Open edX frontend servers, rather than attempt to reconfigure
 > them in-place.
 
+> **Important changes in version 6**
+>
+> As of version 6, this XBlock only supports the Open edX versions `Maple`
+> and higher. Starting with the Maple release, the Open edX community has
+> switched the supported deployment method from edx-configuration playbooks
+> to [Tutor](https://docs.tutor.overhang.io) and the same applies
+> for this XBlock.
+>
+> Instructions for deploying this XBlock with Tutor can be found below,
+> in the Deployment section.
+
 ## Purpose
 
 The hastexo XBlock orchestrates a virtual environment (a "stack") that runs on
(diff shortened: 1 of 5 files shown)


requirements/base.txt
```diff
@@ -19,7 +19,7 @@ tenacity>=6.2,<8
 django<4
 channels>=2.4.0,<3
 asgiref>=3.2.0,<3.2.2  # depends on our channels version
-mysqlclient<2.0.3  # keep in sync with edx-platform
-jsonfield2>=3.0.3,<4   # keep in sync with edx-platform
-pyguacamole>=0.10
+mysqlclient<2.1.0  # keep in sync with edx-platform
+jsonfield>=3.1.0,<4   # keep in sync with edx-platform
```
fghaas, line 23: If this means we can no longer install on Lilac (because it uses a `jsonfield` earlier than 3.1.0), then we need a big bold warning in the changelog about that.

And if we're dropping Lilac compatibility, and can run future releases only Maple, are we sure we need `jsonfield` anymore at all? After all, Django [has a native JSONField, as of 3.1](https://docs.djangoproject.com/en/4.0/releases/3.1/#jsonfield-for-all-supported-database-backends). Does Open edX not use that?

README.md
```diff
@@ -56,6 +67,12 @@ limited by the feature set of the cloud's deployment features.
 
 ## Deployment
 
+# Deployment with Tutor
+
+TODO
```
fghaas, line 72: This is fine but let's please not merge this with this open TODO here. :)

fghaas (changes requested):

fghaas (approved):
This looks good to me now; let's go ahead and merge this. :)

PR #199, opened 2022-02-18
Update the README

mrtmm:
Update the settings example to match the required configuration
for using the `tutor-contrib-hastexo` plugin.


fghaas (changes requested):
Let's please not do this, for reasons outlined in https://github.com/hastexo/tutor-contrib-hastexo/pull/4.

PR #200, opened 2022-02-21
Update the guacd environment variables

mrtmm:
Update the `GUACD_*` environment variables to better suit a
Tutor deployment.
Rename the variables to `GUACD_SERVICE_HOST` and `GUACD_SERVICE_PORT`
to directly read values set for the guacd service with the `tutor k8s`
deployment.
Update the default values to support the `tutor local`/`tutor dev`
deployments.

### DIFF
--- a/hastexo_guacamole_client/consumers.py
+++ b/hastexo_guacamole_client/consumers.py
@@ -20,8 +20,8 @@ class GuacamoleWebSocketConsumer(AsyncWebsocketConsumer):
         """
         Initiate the GuacamoleClient and create a connection to it.
         """
-        guacd_hostname = os.getenv('GUACD_HOSTNAME', 'localhost')
-        guacd_port = os.getenv('GUACD_PORT', 4822)
+        guacd_hostname = os.getenv('GUACD_SERVICE_HOST', 'guacd')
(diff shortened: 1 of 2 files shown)


hastexo_guacamole_client/consumers.py
```diff
@@ -20,8 +20,8 @@ async def connect(self):
         """
         Initiate the GuacamoleClient and create a connection to it.
         """
-        guacd_hostname = os.getenv('GUACD_HOSTNAME', 'localhost')
-        guacd_port = os.getenv('GUACD_PORT', 4822)
+        guacd_hostname = os.getenv('GUACD_SERVICE_HOST', 'guacd')
+        guacd_port = os.getenv('GUACD_SERVICE_PORT', 4822)
```
fghaas, line 24: This is unrelated to the change of variable names, but are we sure that setting the port would ever have worked? Environment variables are strings.

I think you may want to fix this to:

```python
guacd_port = int(os.getenv('GUACD_SERVICE_PORT', '4822'))
```

mrtmm: At the moment I cannot recall when did we use the defaults and not. In any case, it makes sense to correct it, I'll update it now, thanks! :)

fghaas (approved):
Looks good to me now, thanks!

PR #201, opened 2022-02-23
Allow moving on to next provider if one is not configured

mrtmm:
Don't fail to run if one from the list of providers is not
configured. Allow moving on to the next and log a warning message
about failing to initialise a provider.
If none of the listed providers are configured however, log an
error message and raise an exception.

### DIFF
--- a/hastexo/tasks.py
+++ b/hastexo/tasks.py
@@ -160,24 +160,34 @@ class LaunchStackTask(HastexoTask):
         self.providers = []
         for provider in stack.providers:
             p = Provider.init(provider["name"])
-            p.set_logger(logger)
-            p.set_capacity(provider["capacity"])
+            if p:
+                p.set_logger(logger)
+                p.set_capacity(provider["capacity"])
 
-            template = read_from_contentstore(
-                stack.course_id,
-                provider["template"]
-            )
-            p.set_template(template)
-
-            environment_path = provider.get("environment")
-            if environment_path:
-                environment = read_from_contentstore(
+                template = read_from_contentstore(
                     stack.course_id,
-                    environment_path
+                    provider["template"]
                 )
-                p.set_environment(environment)
+                p.set_template(template)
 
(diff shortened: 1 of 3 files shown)


hastexo/tasks.py
```diff
@@ -160,24 +160,34 @@ def run(self, **kwargs):
...
+                        environment_path
+                    )
+                    p.set_environment(environment)
+
+                self.providers.append(p)
+            else:
+                logger.warning(
+                    f'Failed to initialize provider: {provider["name"]}, '
+                    f'make sure the necessary settings are in place or remove '
+                    f'the provider from the list to silence this warning.')
```
fghaas, line 186: Minor note: as far as I know, you can mix f-strings with normal strings in this case, like so:

```python
logger.warning(
    f'Failed to initialize provider: {provider["name"]}, '
    'make sure the necessary settings are in place or remove '
    'the provider from the list to silence this warning.')
```
I tend to prefer that because it tells me at a glance when a part of a string is meant to be substituted, and when it's not.

Also, equally minor: insert comma before "or".

hastexo/tasks.py
```diff
@@ -160,24 +160,34 @@ def run(self, **kwargs):
...
+        if not self.providers:
+            logger.error("No providers were successfully initialised, "
```
fghaas, line 188: We use US spelling for "initialize" earlier, you may want to change this to "initialized".

fghaas (commented):
Excellent! I only have two nits, but other than those this looks great.

hastexo/tasks.py
mrtmm: Ack, done.

hastexo/tasks.py
mrtmm: Good note, updated now, thanks!

fghaas (approved):
Looks great now, thanks!

PR #208, opened 2022-03-17
Add option to track and limit learners lab usage

mrtmm:


hastexo/tasks.py
```diff
@@ -135,6 +135,57 @@ def run(self, **kwargs):
...
 
+        # If a time limit is set for using labs,
+        # check how much time learner has already spent
+        settings = get_xblock_settings()
+        lab_usage_limit = settings.get("lab_usage_limit", None)
+
+        policy_warn_message = None
+        if lab_usage_limit:
+            # don't look further than 365 days for lab launch
+            cutoff = timezone.now() - timezone.timedelta(days=365)
```
fghaas, line 146: Future implementation note, no action needed now we probably want to make this window configurable as `lab_usage_limit_window`.

hastexo/models.py
```diff
@@ -35,6 +35,9 @@ class Meta:
     name = models.CharField(max_length=64, db_index=True)
     student_id = models.CharField(max_length=40, db_index=True)
     course_id = models.CharField(max_length=50, db_index=True)
+    learner = models.ForeignKey(django_settings.AUTH_USER_MODEL,
```
fghaas, line 38: I'm still a bit on the fence about this being called `learner`, as opposed to calling this `user` and renaming the existing `user` field to `username`.

hastexo/tasks.py
```diff
@@ -135,6 +135,57 @@ def run(self, **kwargs):
...
+                if policy == 'block':
+                    raise Exception(
+                        "You've reached the time limit allocated for you "
+                        "for using labs.")
+
+                elif policy == 'warn':
+                    policy_warn_message = (
+                        "You've reached the time limit allocated for you "
+                        "for using labs.")
+
```
fghaas, line 188: This is missing the `notify` option mentioned in the README. It will either need to be implemented here (essentially, that's a log message using the `django` logger, see [the documentation](https://docs.djangoproject.com/en/4.0/ref/logging/#default-logging-configuration)), or removed from the README. :)

fghaas (commented):
This looks good to me! I have a few notes, and maybe @foadlind is also inclined to take an extra look. :)

hastexo/tasks.py
fghaas: Yeah you can totally include that in a follow-up change too though.

mrtmm: Ack, I removed the reference to it from the README for the time being.

hastexo/models.py
mrtmm: As discussed elsewhere, we will stick with the `learner` field.

hastexo/tasks.py
mrtmm: Ack.

hastexo/tasks.py
```diff
@@ -135,6 +135,57 @@ def run(self, **kwargs):
...
+                policy = settings.get("lab_usage_limit_breach_policy",
+                                      "").lower()
+
+                if policy:
+                    logger.error(
+                        f'Learner {stacks[0].learner.email} has gone over '
+                        'the lab usage limit!')
+
+                if policy == 'block':
+                    raise Exception(
```
fghaas, line 180: I'm not sure how I feel about this being a generic Exception and being displayed as such to the learner; the fact that the pop-up reads `Exception("<message>")` isn't particularly pretty. I'm not sure what the best option here; should this perhaps be a subclass of `LaunchStackFailed`?

hastexo/tasks.py
```diff
@@ -135,6 +135,57 @@ def run(self, **kwargs):
...
+                        "You've reached the time limit allocated for you "
```
fghaas, line 181: Should be "allocated to you".

fghaas (commented):
Functionally I think this looks good now, and we could merge this to `main` as-is. I have a couple of notes about user-facing error messages; see the inline comments for those.

hastexo/tasks.py
mrtmm: Ack.

hastexo/tasks.py
```diff
@@ -135,6 +135,57 @@ def run(self, **kwargs):
...
+                        "for using labs.")
+
+                elif policy == 'warn':
+                    policy_warn_message = (
+                        "You've reached the time limit allocated for you "
```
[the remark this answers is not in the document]
mrtmm: Ack.

hastexo/tasks.py
mrtmm: Right, yes, I started out with `LaunchStackFailed` but ran into some errors how javascript handled it and switched to the generic one for testing. I'll take another look at that now :)

fghaas: One more thought on this: when blocked, the stack `status` ends up being `LAUNCH_ERROR`. I guess that's OK, we just need to make sure that such stacks get correctly reaped at the end of their `delete_age`. Since `LAUNCH_ERROR` is _not_ in the `dont_delete` list in [ReaperJob](https://github.com/hastexo/hastexo-xblock/blob/master/hastexo/jobs.py#L112) I would expect this to behave correctly and no changes are needed; do you agree?

fghaas (commented):
I think this is good to merge to `master` now. @foadlind, any objections on something I missed?

README.md
```diff
@@ -325,6 +325,19 @@ This is a brief explanation of each:
...
 
+* `lab_usage_limit`: Allocate limited time per user for labs across the platform,
+  in seconds. (Default is `None`, meaning there is no time limit).
+
+* `lab_usage_limit_breach_policy`: What to do when the learner's lab
+  limit has been exceeded.
+
+  * `None` (the default) means just write a message to the logs.
+
+  * `warn` will display a display a warning to the learner that the
```
foadlind, line 336: "display a" is repeated twice.

foadlind (commented):
Looks good to me. I only have a small nit :)

README.md
fghaas: I'm pretty sure I injected that. Sorry!

fghaas (approved):
Looks good to me now, thank you!

PR #210, opened 2022-04-12
Add specific exception case for restricting lab access

mrtmm:
Add new Exception class to be more specific when raising the exception
for restricting lab access due to lab usage limit being reached.


fghaas:
Looks good to me, just fix the two micro-nits in the commit message please ("exeption", "forrestricting"). :)

fghaas (approved):
Looks good to me now, thanks!

PR #215, opened 2022-06-09
Fix YAML syntax for HASTEXO_XBLOCK_SETTINGS in README

mrgnr:
The example YAML for `HASTEXO_XBLOCK_SETTINGS` is not well-formed in the README. If I paste the example directly into my `config.yml`, I get errors running Tutor because it cannot parse the file.

## Done

- Removed commas between dictionary items

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,31 +91,31 @@ two steps:
     launch_timeout: 900
     providers:
       default:
-        type: openstack,
-        os_auth_url: "",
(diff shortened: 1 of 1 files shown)


fghaas (approved):
This is correct. The commas must have been an ancient copy-and-paste leftover from when Open edX configuration files converted from a default JSON to a default YAML syntax. Thanks @mrgnr!

PR #217, opened 2022-06-22
Bump Paramiko upper bound to 2.11

fghaas:
Paramiko versions prior to 2.10.1 include a race condition that may allow information disclosure. Bump the upper bound to 2.11 to use a newer Paramiko.

Prior to dfd06ddebcf35e71891e9f2aa688899cdf435401, we already did install the newer Paramiko release (inadvertently) when installing the package with pip. With this, we also install the correct version when running `pip install -r requirements.txt`.


fghaas:
@mrtmm One more small change before I cut 6.1.2.

mrtmm (approved):

PR #221, opened 2022-06-29
Add a check to the `add_user_foreign_key` migration

mrtmm:
It may happen that there are some problematic stacks in the database
that would cause the migration to fail when the stacks cannot be
linked to a user account. These stacks are most likely a result of
testing with staff access, overriding some requirements that learner
stacks are subject to.

Add a check to the migration to find out if such stacks exist and
if they do, raise an exception with a message providing the list of
problematic stack names and guidance how to proceed.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,3 +1,12 @@
+Unreleased
+-------------------------
+
+* [Bug fix] Add a check to the `add_user_foreign_key` migration
+  file to find stacks that are missing a link to a real user account.
+  If such stack(s) exist, do not attempt to apply the migration,
+  instead raise an exception and provide an error message with
+  guidance on how to proceed.
+
 Version 6.1.2 (2022-06-22)
 -------------------------
 
--- a/hastexo/migrations/0010_add_user_foreign_key.py
+++ b/hastexo/migrations/0010_add_user_foreign_key.py
@@ -1,9 +1,30 @@
 from django.conf import settings
-from django.db import migrations, models
+from django.db import migrations, models, OperationalError
 import django.db.models.deletion
 
 
 class Migration(migrations.Migration):
+
+    def check_stacks(apps, schema_editor):
+        """
+        Check if all stacks can be linked to a real user account.
+        If not, error out with a suggestion how to proceed.
+        """
+        Stack = apps.get_model("hastexo", "Stack")
+        AnonymousUserId = apps.get_model("student", "AnonymousUserId")
+        anonymous_user_ids = AnonymousUserId.objects.values(
+            'anonymous_user_id').distinct()
+        problematic_stacks = Stack.objects.exclude(
+            student_id__in=anonymous_user_ids)
+
+        if len(problematic_stacks) > 0:
+            raise OperationalError(
+                'Unable to link stacks to users; please make sure that '
+                'the following stacks have a student_id that corresponds '
+                'to a real user account: '
+                f'{[s.name for s in problematic_stacks]}. '
+                'Please update or delete the stacks manually and rerun '
+                'the migration.')
 
     def backfill_learner(apps, schema_editor):
         """
@@ -22,6 +43,7 @@ class Migration(migrations.Migration):
     ]
 
     operations = [
+        migrations.RunPython(check_stacks),
         migrations.AddField(
             model_name='stack',
             name='learner',


Changelog.md
```diff
@@ -1,3 +1,12 @@
+Unreleased
+-------------------------
+
+* [Bug fix] Add a check to the `add_user_forgeign_key` migration
```
fghaas, line 4: Typo, should be "foreign" (same in the commit message and PR description).

hastexo/migrations/0010_add_user_foreign_key.py
```diff
@@ -1,10 +1,29 @@
...
 import django.db.models.deletion
 
 
 class Migration(migrations.Migration):
 
+    def check_stacks(apps, schema_editor):
+        Stack = apps.get_model("hastexo", "Stack")
+        AnonymousUserId = apps.get_model("student", "AnonymousUserId")
+        stacks_with_no_user = []
+        for stack in Stack.objects.all():
```
fghaas, line 13: Ok so this is a loop over _all_ `Stack` objects with a nested `get` on another object type. I worry that in an environment with thousands of stacks and hundreds of users this would take forever. I do appreciate that this is a migration so we might not worry about such things in general, but here I'm guessing we might actually hit standard 30-minute CI pipeline timeouts. So if this is the only way to do it then let's keep this, but if we _can_ make this faster, we should. 

Is there any chance this could be rewritten as a [filter query](https://docs.djangoproject.com/en/4.0/topics/db/queries/#retrieving-specific-objects-with-filters)?

hastexo/migrations/0010_add_user_foreign_key.py
```diff
@@ -1,10 +1,29 @@
...
+            try:
+                stack.learner = AnonymousUserId.objects.get(
+                    anonymous_user_id=stack.student_id).user
+            except ObjectDoesNotExist:
+                stacks_with_no_user.append(stack.name)
+        if len(stacks_with_no_user) > 0:
+            raise OperationalError(
+                'Unable to link stacks to users; please make sure that '
+                'the following stacks have a student_id that corresponds '
+                f'to a real user account: {stacks_with_no_user}.'
```
fghaas, line 23: Micro-nit: this needs whitespace after the the full-stop.

Changelog.md
mrtmm: Thanks, fixed.

hastexo/migrations/0010_add_user_foreign_key.py
mrtmm: Thanks, done.

hastexo/migrations/0010_add_user_foreign_key.py
mrtmm: I believe we can use exclude, I've updated the patch just now, please have a look.

hastexo/migrations/0010_add_user_foreign_key.py
```diff
@@ -1,10 +1,30 @@
...
+    def check_stacks(apps, schema_editor):
+        """
+        Check if all stacks can be linked to a real user account.
+        If not, error out with a suggestion how to proceed.
+        """
+        Stack = apps.get_model("hastexo", "Stack")
+        AnonymousUserId = apps.get_model("student", "AnonymousUserId")
+        problematic_stacks = Stack.objects.exclude(
+            student_id__in=AnonymousUserId.objects.all().values(
+                'anonymous_user_id'))
```
fghaas, line 17: That's a good approach. I *think* you can rewrite this such that the lazily evaluated `QuerySet` becomes even more efficient, like so (note that this is untested):

```python
anonymous_user_ids = AnonymousUserId.objects.values('anonymous_user_id').distinct()
problematic_stacks = Stack.objects.exclude(student_id__in=anonymous_user_ids)
```

(From memory, `objects.values()` is the same as `objects.all().values()`)

foadlind: I don't think assigning it to a variable will cause the first queryset to be evaluated. The ORM will still create one big query from these two lines. 
You can change that to:
```python
anonymous_user_ids = list(AnonymousUserId.objects.values('anonymous_user_id').distinct())
problematic_stacks = Stack.objects.exclude(student_id__in=anonymous_user_ids)
```

mrtmm: Ack, I'll run my local test steps with this and update the patch.

mrtmm: using `list()` resulted in all (or at least a very large number of) stacks being excluded, I updated the patch without it, using @fghaas example.

fghaas (approved):

PR #223, opened 2022-07-01
Attempt to fix the add_user_foreign_key migration

fghaas:
I have no idea if this actually works. :) But it seemed easier to just hack this up than to try to describe it. :)

### DIFF
--- a/hastexo/migrations/0010_add_user_foreign_key.py
+++ b/hastexo/migrations/0010_add_user_foreign_key.py
@@ -1,30 +1,13 @@
 from django.conf import settings
-from django.db import migrations, models, OperationalError
+from django.core.exceptions import ObjectDoesNotExist
+from django.db import migrations, models
 import django.db.models.deletion
+
+import logging
+logger = logging.getLogger(__name__)
 
 
 class Migration(migrations.Migration):
-
-    def check_stacks(apps, schema_editor):
-        """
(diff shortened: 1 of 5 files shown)


hastexo/migrations/0010_add_user_foreign_key.py
```diff
@@ -33,17 +15,22 @@ def backfill_learner(apps, schema_editor):
         Stack = apps.get_model("hastexo", "Stack")
         AnonymousUserId = apps.get_model("student", "AnonymousUserId")
         for stack in Stack.objects.all():
-            stack.learner = AnonymousUserId.objects.get(
+            learner = AnonymousUserId.objects.get(
```
mrtmm, line 18: I believe this will throw `DoesNotExist` instead of returning None if user cannot be found

fghaas: OK, [like so](https://github.com/hastexo/hastexo-xblock/compare/4ef5ecfb389bc16f65264a4f001553e046a136eb..b50fa24bd46081c0b8de450c277db01839c65d41)? (Reference: [Django docs](https://docs.djangoproject.com/en/3.2/ref/exceptions/#objectdoesnotexist))

mrtmm: Yes, that should do it.

mrtmm (approved):

PR #226, opened 2022-07-06
WIP: Require asgiref<3.4

fghaas:
Alternate approach to #225, just to get some test results.


fghaas:
This is likely to _not_ work, as the issue we ran into is most likely present from 3.3.0 forward.

PR #228, opened 2022-07-13
Apply additional fix to add_user_foreign_key migration

fghaas:
The hack in 583fb729b1e201c830579345dca5beca4b131006 modified
0010_add_user_foreign_key in such a way that it ended up *not* setting
a database constraint when it should have.

Enable the database-enforced constraint in the right place.


fghaas:
@mrtmm: _Maybe_ this fixes our issue; I'm not sure at all. :/

fghaas:
@mrtmm, I'm suddenly not sure if I didn't do an accidental force-push that clobbers over your change. Can you visually compare this to your local checkout before you approve please?

fghaas:
@mrtmm Never mind, I compared MD5 checksums for all the migration modules and we're fine here. This is good to merge, as far as I am concerned. :)

PR #230, opened 2022-07-20
Add an option to override the global `suspend_timeout` setting

mrtmm:
Introduce a new XBlock attribute for overriding the `suspend_timeout`
setting.

### DIFF
--- a/hastexo/jobs.py
+++ b/hastexo/jobs.py
@@ -66,15 +66,33 @@ class SuspenderJob(AbstractJob):
 
         # Get stacks to suspend
         with transaction.atomic():
-            stacks = Stack.objects.select_for_update().filter(
+            stacks_without_suspend_by = Stack.objects.select_for_update().filter(  # noqa: E501
                 suspend_timestamp__isnull=False
+            ).filter(
+                suspend_by__isnull=True
             ).filter(
                 suspend_timestamp__lt=cutoff
             ).filter(
                 status__in=states
             ).exclude(
                 provider__exact=''
-            ).order_by('suspend_timestamp')[:concurrency]
+            )
+
+            stacks_with_suspend_by = Stack.objects.select_for_update().filter(
+                suspend_timestamp__isnull=False
+            ).filter(
+                suspend_by__isnull=False
+            ).filter(
+                suspend_by__lt=timezone.now()
+            ).filter(
+                status__in=states
+            ).exclude(
+                provider__exact=''
+            )
+
+            stacks = stacks_without_suspend_by.union(
+                stacks_with_suspend_by).order_by(
+                    'suspend_timestamp')[:concurrency]
 
             for stack in stacks:
                 stack.status = SUSPEND_PENDING
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,3 +1,9 @@
+Unreleased
+-------------------------
+
+* [Enhancement] Add an option to override the `suspend_timeout`
+  global setting via an XBlock attribute, in seconds.
+
 Version 6.1.5 (2022-07-13)
 -------------------------
 
--- a/README.md
+++ b/README.md
@@ -623,6 +623,10 @@ The following are optional:
 * `launch_timeout`: How long to wait for a stack to be launched, in seconds.
   If unset, the global timeout will be used.
 
+* `suspend_timeout`: Timeout for how long to wait before suspending a stack,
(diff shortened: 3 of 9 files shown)


hastexo/jobs.py
```diff
@@ -69,7 +66,9 @@ def run(self):
             stacks = Stack.objects.select_for_update().filter(
                 suspend_timestamp__isnull=False
             ).filter(
-                suspend_timestamp__lt=cutoff
+                suspend_by__isnull=False
+            ).filter(
+                suspend_by__lt=timezone.now()
```
fghaas, line 71: Hmmm maybe I'm missing something, but we don't backfill the `suspend_by` field, correct? 

So, if 
* we're not backfilling but 
* we are looking only for objects where `suspend_by` is not null, and where its content is a date in the past, 
 
doesn't that mean we effectively *don't* suspend the labs that have ceased sending keepalives less than `suspend_timeout` ago, i.e. that were last touched in the two minutes prior to migration, until the learner opens them again?

mrtmm: Good point, I updated the PR.

hastexo/jobs.py
```diff
@@ -66,15 +66,32 @@ def run(self):
 
         # Get stacks to suspend
         with transaction.atomic():
-            stacks = Stack.objects.select_for_update().filter(
+            stacks_1 = Stack.objects.select_for_update().filter(
```
fghaas, line 69: To make this easier to understand and more readable, I'd suggest to call this variable `stacks_without_suspend_by`.

hastexo/jobs.py
```diff
@@ -66,15 +66,32 @@ def run(self):
...
             ).filter(
                 suspend_timestamp__lt=cutoff
             ).filter(
                 status__in=states
             ).exclude(
                 provider__exact=''
-            ).order_by('suspend_timestamp')[:concurrency]
+            )
+
+            stacks_2 = Stack.objects.select_for_update().filter(
```
fghaas, line 81: ... and this, `stacks_with_suspend_by`.

hastexo/jobs.py
```diff
@@ -66,15 +66,32 @@ def run(self):
...
+            ).filter(
+                suspend_by__lt=timezone.now()
+            ).filter(
+                status__in=states
+            ).exclude(
+                provider__exact=''
+            )
+
+            stacks = stacks_1.union(stacks_2).order_by(
+                'suspend_timestamp')[:concurrency]
```
fghaas, line 94: I'm not quite sure what `[:concurrency]` does here, could you fill me in please?

fghaas (commented):
This looks good to me now, modulo one style (naming) note and a question.

Just checking to see if I understand this correctly: if a learner comes to one lab where the course author has set `suspend_timeout` to, say, 1800, and then they move on to another lab where the XBlock has no `suspend_timeout` attribute, then as they move away from the lab the default 120s timeout applies. Correct?

hastexo/jobs.py
mrtmm: Ack.

hastexo/jobs.py
mrtmm: Ack.

hastexo/jobs.py
mrtmm: It limits the number of stacks to suspend in one job run.

mrtmm (commented):
>  if a learner comes to one lab where the course author has set suspend_timeout to, say, 1800, and then they move on to another lab where the XBlock has no suspend_timeout attribute, then as they move away from the lab the default 120s timeout applies. Correct?
Yes.

hastexo/jobs.py
fghaas: Ah! 💡 

Thanks!

hastexo/jobs.py
```diff
@@ -66,15 +66,33 @@ def run(self):
 
         # Get stacks to suspend
         with transaction.atomic():
-            stacks = Stack.objects.select_for_update().filter(
+            stacks_without_suspend_by = Stack.objects.select_for_update().filter(
```
fghaas, line 69: Whoops, now `flake8` chokes on this line. :) Might want to add `# noqa: E501` at the end rather than break this line.

fghaas (approved):
This looks good to me now. If the tests pass I'll go ahead and merge this.

PR #233, opened 2022-08-12
Update the `hastexo_guacamole_client` to Channels 3

mrtmm:
Following the release notes [1] and Channels documentation [2] for
upgrading our `hastexo_guacamole_client` to Channels 3:
* replace routing.py with asgi.py, with some changes to the asgi application definition,
  including using the new consumer `as_asgi()` class method
  and implementing the AllowedHostsOriginValidator.
* explicitly set our `sync_to_async` calls to the GuacamoleClient`thread_sensitive=False`,
  as starting from asgiref==3.3.0 (requirement for Channels 3),
  the default is thread_sensitive=True [3] which tries to run everything in the same thread.
  For us the result is that the LMS will start receiving information from the guacamole
  client just fine at first but any concurrent connection will be blocked by the first,
  still open, connection.

[1] https://channels.readthedocs.io/en/stable/releases/3.0.0.html
[2] https://channels.readthedocs.io/en/stable/topics/consumers.html#asyncwebsocketconsumer
[3] https://github.com/django/asgiref/commit/7becc9daca2628c46af1cb7e46b4c47c1ea27adf


hastexo_guacamole_client/consumers.py
```diff
@@ -66,7 +66,7 @@ async def disconnect(self, code):
         Close the GuacamoleClient connection on WebSocket disconnect.
         """
         self.task.cancel()
-        await sync_to_async(self.client.close)()
+        await sync_to_async(self.client.close, thread_sensitive=False)()
```
foadlind, line 69: I suggest adding a comment here to document why we want this to be `False` and not the default `True`.

hastexo_guacamole_client/consumers.py
```diff
@@ -84,6 +84,7 @@ async def open(self):
         Receive data from GuacamoleClient and pass it to the WebSocket
         """
         while True:
-            content = await sync_to_async(self.client.receive)()
+            content = await sync_to_async(
+                self.client.receive, thread_sensitive=False)()
```
foadlind, line 88: Same suggestion as above.

foadlind (commented):
Thanks @mrtmm, it looks good. I only have a small suggestion. See the inline comments.

hastexo_guacamole_client/consumers.py
mrtmm: Ack, done.

hastexo_guacamole_client/consumers.py
mrtmm: Done :)

foadlind (dismissed):
Looks good now. Thanks!

requirements/base.txt
```diff
@@ -16,9 +16,9 @@ python-novaclient>=7.1.2,<16
 tenacity>=6.2,<8
 
 # for hastexo_guacamole_client
-django<3.2
-channels>=2.4.0,<3
-asgiref~=3.2.0 # depends on our channels version
+django~=3.2.1
+channels>2.4.0,<=3.0.5
```
fghaas, line 20: I think this is functionally correct (there has been no Channels 2.x release since 2.4.0), but just to make it clearer you may want to set this to `channels>=3.0.0,<=3.0.5` or perhaps `channels~=3.0.0` which more clearly reflects that this depends on the Channels 3.x.x major release.

fghaas (commented):
Looks excellent to me too! I just have a tiny suggestion for `requirements.txt`.

requirements/base.txt
mrtmm: Ack, done. I went with `channels>=3.0.0,<=3.0.5` because I think it's better if we bump the upper requirement for channels and asgiref consciously instead of allowing a newer version being pulled in once released without testing it before.

fghaas (approved):
Just re-approving this one since @foadlind's review was auto-dismissed by the last force-push. :)

PR #235, opened 2022-08-18
Manually register Celery class-based tasks

foadlind:
From Celery 5.0.0 the [legacy task API was discontinued](https://github.com/celery/celery/pull/6315), which means that the Task base class no longer automatically registers child tasks.
Therefore to use hastexo-xblock with Open edX Nutmeg (which uses Celery 5.2.6), we need to [manually register the class-based tasks](https://docs.celeryq.dev/en/master/history/whatsnew-4.0.html#the-task-base-class-no-longer-automatically-register-tasks) on the Celery app instance.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,6 +1,10 @@
 Unreleased
 -------------------------
-
+* [Bug fix] From Celery 5.0.0 the legacy task API was discontinued.
+  This meant that the Task base class no longer automatically registered 
+  child tasks in Open edX Nutmeg (which uses Celery 5.2.6). 
+  Manually register the class-based tasks on the Celery app instance.
+  
 * [BREAKING CHANGE] Update the `hastexo_guacamole_client` to
   Channels 3. The asgi root application (`ASGI_APPLICATION`)
   is now defined in the `asgi.py` file instead of `routing.py` file.
--- a/hastexo/hastexo.py
+++ b/hastexo/hastexo.py
@@ -738,7 +738,7 @@ class HastexoXBlock(XBlock,
         soft_time_limit = self.get_launch_timeout(settings)
         hard_time_limit = soft_time_limit + 30
 
-        return LaunchStackTask().apply_async(
+        return LaunchStackTask.apply_async(
             kwargs=kwargs,
             expires=soft_time_limit,
             soft_time_limit=soft_time_limit,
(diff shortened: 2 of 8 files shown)


foadlind:
Yes, the new Task API was available in Celery 4. I haven't tested this change with Maple yet, since my local Tutor environment is on Nutmeg now.

mrtmm:
Looks good to me as well! I would perhaps not call it a bug fix as upgrading/adding support to a newer version of something is rather normal course of events but I'll trust your judgement on that :)

foadlind:
Why is codecov blocking this? I am not able to see the detail report. I get a 403 when I click on the "Details" button of `codecov/patch`. Do I have to sign up to see the report?

foadlind:
> I don't think you have to sign up, just signing in with your GitHub account should suffice.
> 
> That said, the way I read https://app.codecov.io/gh/hastexo/hastexo-xblock/compare/235 the tests actually do hit 100% of the diff, so I really don't know where the block comes from. It's definitely not stopping us from merging, if the other tests pass (which they now do).

Today I didn't even need to sign in. It just worked. :man_shrugging:
Ok so it is not blocking but as far as I understand the coverage report says that the code I changed (the diff) is only covered 75% while on the whole the change has increased the coverage of the codebase by 0.17%. And that is not ok because the target for a diff coverage is 94.93%. For such a scientific measure, I hope it is not counting the lines I added to the Changelog :laughing:

foadlind:
> One other thing though: your patch doesn't touch requirements/test.txt, so I think what might have happened here is that the CI tests ran with Celery 4. Do we want that?

That's a good point. I'll increase the upper bound to Celery <=  5.2.6. I think it's good to keep Celery 4 lower bound in case we want to port this back to Maple.

fghaas:
Right, one other option might be to test with both Celery 4 and 5; see https://github.com/foadlind/hastexo-xblock/pull/1 for that.

foadlind:
Right, one other option might be to test with both Celery 4 and 5; see https://github.com/foadlind/hastexo-xblock/pull/1 for that.
That's even better. Thanks! I'll add that in here.

fghaas:
Excellent, I think we have all the bases covered then. @mrtmm, do you agree that this is good to merge?

mrtmm (approved):
Yes, looks good to me, thanks!

fghaas:
And done! Tests take twice as long now, obviously, but hey — that's what a CI pipeline is for. :)

foadlind:
Thanks for your patience and reviews! :slightly_smiling_face:

PR #241, opened 2022-09-30
Retry `read_from_contentstore()`

mrtmm:
There might occasionally be moments where we cannot reach MongoDB on first try.
Implement `tenacity`'s retry functionality for the purpose of trying again when that happens.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,3 +1,8 @@
+Unreleased
+----------
+* [Bug fix] Retry `read_from_contentstore()`. Use `tenacity`'s
+  retry functionality for getting course information from contentstore.
+
 Version 7.0.0 (2022-08-22)
 -------------------------
 * [Bug fix] From Celery 5.0.0 the legacy task API was discontinued.
--- a/hastexo/common.py
+++ b/hastexo/common.py
@@ -12,6 +12,14 @@ from paramiko.ssh_exception import (AuthenticationException,
                                     SSHException,
                                     NoValidConnectionsError)
 from django.conf import settings as django_settings
+from pymongo.errors import PyMongoError
+from tenacity import (
+    retry,
+    retry_if_exception_type,
+    stop_after_attempt,
+    wait_exponential,
+    before_sleep_log,
+)
 
 from .models import Stack
 
@@ -238,6 +246,15 @@ def get_stack(name, course_id, student_id, prop=None):
         return stack
 
 
+# If PyMongoError is raised here, try again (max attempts = 3)
+# Before every subsequent retry, wait
+# Use before_sleep_log to log attempts
+# Reraise the exception from last attempt
+@retry(retry=retry_if_exception_type(PyMongoError),
+       stop=stop_after_attempt(3),
+       wait=wait_exponential(),
+       before_sleep=before_sleep_log(logger, logging.WARNING),
+       reraise=True)
 def read_from_contentstore(course_key, path):
     """
     Loads a file directly from the course's content store.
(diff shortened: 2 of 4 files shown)


hastexo/common.py
```diff
@@ -238,6 +246,15 @@ def get_stack(name, course_id, student_id, prop=None):
...
 
+# If ServerSelectionTimeoutError is raised here, try again (max attempts = 3)
+# Before every subsequent retry, wait
+# Use before_sleep_log to log attempts
+# Reraise the exception from last attempt
+@retry(retry=retry_if_exception_type(ServerSelectionTimeoutError),
+       stop=stop_after_attempt(3),
+       wait=wait_exponential(),
+       before_sleep=before_sleep_log(logger, logging.WARNING),
+       reraise=True)
```
fghaas, line 257: Do I understand correctly that the default initial base with `wait_exponential()` is 2, so this means that we'll retry after 1 (2⁰) seconds, then 2 (2¹), then 4 (2²), and if it still doesn't succeed we give up?

fghaas: Also, do we only want to trigger the retry on `ServerSelectionTimeoutError`? Or do we perhaps want to expand this to all `PyMongoError` exceptions? (It's a read operation; redoing it in case of an error should kind of not hurt for any reason, unless I am missing something.)

Changelog.md
```diff
@@ -1,3 +1,8 @@
+Unreleased
+----------
+* [Enhancement] Retry `read_from_contentstore()`. Use `tenacity`'s
```
fghaas, line 3: I'd say making the code more robust is a bug fix, meaning I'd bump the patch number rather than the minor here. Would you agree?

fghaas (commented):
This looks good! I have just a few questions.

fghaas:
Oh, one more thing, just for the record: we'll need a couple of tests please, where we're mocking the pymongo errors. But I'm sure you have that on your radar already. :)

Changelog.md
mrtmm: Ack, makes sense.

tests/unit/test_common.py
```diff
@@ -71,6 +72,59 @@ def test_read_from_contentstore(self):
...
+
+        # Assert
+        self.assertEqual(content, "bogus")
+        self.assertEqual(contentstore_mock.return_value.find.call_count, 3)
+
+    def test_read_from_contentstore_retry_fail(self):
+        # Setup
+        contentstore_mock = Mock()
+        contentstore_mock.return_value.find.side_effect = [
+            PyMongoError,
```
fghaas: Small suggestion: when code catches the abstract superclass of an exception category (in our case, `PyMongoError`), then I think it's better to write the side effects in the test code so that they throw concrete subclasses.

I've pushed a fixup commit doing that in 19f12a7924135f3fe40edbd395b5c149cf3c44ea, please see if that makes sense to you.

fghaas (commented):
Looks good to me now; I just have one more small suggestion about the tests.

tests/unit/test_common.py
fghaas: OK, I've squashed the fixup commit and force-pushed to your branch. I think this is good to merge, then.

mrtmm: Ack, thanks!

fghaas:
Huh? Why is the test failing _now,_ after a change to just the commit message? Makes no sense. I'll try rerunning it.

fghaas (commented):
Tests completed after a simple retry. This is good to merge.

fghaas (approved):

PR #248, opened 2022-12-02
Add internationalization support

mrtmm:


hastexo/static/html/main.html
```diff
@@ -28,15 +30,15 @@ <h1 class="header">Sorry!</h1>
...
+      <p class="message">{% trans "Working" %}</p>
     </div>
   </div>
   <div class="dialog" id="check_complete">
     <div class="inner">
-      <h1 class="check_result_heading">Progress check result</h1>
-      <p class="check_result_message">You completed <strong class="check_pass">0</strong> out of <strong class="check_total">0</strong> tasks.</p>
-      <h2 class="hints_title">Hints</h2>
+      <h1 class="check_result_heading">{% trans "Progress check result" %}</h1>
+      <p class="check_result_message">{% trans "You completed" %} <strong class="check_pass">0</strong> {% trans "out of" %} <strong class="check_total">0</strong> {% trans "tasks." %}</p>
```
fghaas, line 40: I'm still not quite sure if this is the best way to do this. I think you might be better off using [`trans` with placeholders](https://jinja.palletsprojects.com/en/3.0.x/templates/#i18n) here. Would you agree?

fghaas (commented):
This generally looks very good to me. I have one minor suggestion though.

hastexo/static/html/main.html
mrtmm: I agree that this is probably not the best way but we can't use those placeholders here because we are not passing the values in the context when we render the template, we are attaching them in javascript. I tried to build it as one string and get it translated in js but all my attempts failed so far. But, I just might have figured where I went wrong, I'll try again to tweak the js code.

mrtmm: I updated the PR now, I defined the string in js, translate with placeholders and then replace the values after translation is loaded. Should be better.

fghaas (approved):
This looks good to me now! I'd say we can merge this as soon as we've verified that the tests are now passing.

PR #249, opened 2023-01-20
Fix unit and integration tests (and GitHub Actions workflows)

fghaas:
tox 4.0.0, released in December 2022, removed the `tox.venv` module which `tox-pip-version` relies on.

Therefore, until
    
* tox-pip-version starts supporting tox 4, or
* we decide that we no longer need to test with multiple pip versions, and thus no longer have a need for installing tox-pip-version,
    
force a `tox` version earlier than 4.

Also, update the GitHub actions to their latest major releases, and drop the pip-installed `mock` module for the standard library's `unittest.mock` (like we should have done years ago, really).

### DIFF
--- a/.github/workflows/tox.yml
+++ b/.github/workflows/tox.yml
@@ -25,7 +25,7 @@ jobs:
         python-version: ${{ matrix.python-version }}
     - name: Install dependencies
       run: |
-        pip install tox tox-gh-actions tox-pip-version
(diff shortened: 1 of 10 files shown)


fghaas:
@mrtmm If we merge this, and rebase #248 on it, then your unit tests should succeed both locally and in the GitHub Actions workflow. :)

mrtmm (approved):

PR #251, opened 2023-02-17
Copy&Paste to and from the terminal

mrtmm:
* Restore the possibility to paste into the terminal by updating our use of the `guacamole-js` library. We cannot, at this time, add support for the paste functionality via the Async Clipboard API due to the problematic support for the API by different browsers.
* Add (partial) support for copying text out from the terminal via the Async Clipboard API. Partial support, again, due to the problematic support for the API by different browsers. This will work with browsers that support the API and don't block access to it.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,3 +1,12 @@
+Unreleased
+-------------------------
+* [Bug fix] Restore the `paste` functionality by addressing
+  the related changes in the `guacamole-js` library.
+* [Enhancement] Add support for copying text out from
+  the terminal using the Async Clipboard API.
+  This works on Mozilla Firefox at this time. Support
+  in other browsers may follow.
+
 Version 7.3.0 (2023-01-20)
 -------------------------
 * [Enhancement] Add internationalization support.
--- a/README.md
+++ b/README.md
@@ -660,6 +660,43 @@ stack is resumed automatically.  They are connected to the same training
 environment they were working with before, in the *same state* they left it in.
 (The process of suspension works just like in a home computer.)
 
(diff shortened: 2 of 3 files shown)


Changelog.md
```diff
@@ -1,3 +1,11 @@
+Unreleased
+-------------------------
+* [Bug fix] Restore the `paste` functionality by addressing
+  the related changes in the `guacamole-js` library.
+* [Enhacement] Add support for copying text out from
```
fghaas, line 5: Typo, should be "Enhancement". :)

Changelog.md
```diff
@@ -1,3 +1,11 @@
...
+  the terminal using the Async Clipboard API. Partial, due
+  to the lack of support for the API by different browsers.
```
fghaas, line 7: I think the changelog should say what we know about which browsers work and which do not. Assuming we know this to work only on Firefox at this point, I'd put:

```suggestion
  the terminal using the Async Clipboard API.
  This works on Mozilla Firefox at this time. Support in other browsers may follow.
```

fghaas (commented):
This looks great! I only have a minor comment about the changelog.

Also, do we want to explain copy and paste in the README? We haven't done so before but maybe now is a good time to spell it out?

Changelog.md
mrtmm: Thanks :)

Changelog.md
mrtmm: Ack, updated :)

mrtmm:
Yes, makes sense, I tried to add some explanation to the README.

PR #253, opened 2023-03-01
Update requirements to support Open edX Olive release

mrtmm:


Changelog.md
```diff
@@ -1,3 +1,7 @@
+Unreleased
+-------------------------
+* [Enhancement] Update requirements for Open EdX Olive release.
```
fghaas, line 3: Micro-nit:

```suggestion
* [Enhancement] Update requirements for Open edX Olive release.
```

fghaas (commented):
If I read this correctly, then there are no breaking version conflicts here so this should install fine on both Nutmeg and Olive, which is excellent!

I just have one thing that's too tiny to even be called a nit. :)

fghaas (approved):
This looks good to me!

PR #257, opened 2023-08-16
Add support for Ed25519 SSH keys

mrtmm:
Add a new xblock attribute that allows to define an ssh key type for a lab environment.

### DIFF
--- a/README.md
+++ b/README.md
@@ -482,6 +482,11 @@ The following are optional:
 * `stack_template_path`: The static asset path to the orchestration template,
   if not specified per provider below.
 
+* `stack_key_type`: An SSH key type for accessing the lab environment.
+  Options are `rsa`, `ed25519` and `None` (default). If a key type is chosen,
+  a key with the selected type will be generated for the lab environment.
+  If set to `None`, the key handling should be done via the lab template.
+
 * `launch_timeout`: How long to wait for a stack to be launched, in seconds.
   If unset, the global timeout will be used.
 
--- a/hastexo/common.py
+++ b/hastexo/common.py
@@ -8,6 +8,7 @@ import six
 
 from io import StringIO
 from socket import timeout as SocketTimeout
+from paramiko import RSAKey, Ed25519Key
 from paramiko.ssh_exception import (AuthenticationException,
                                     SSHException,
                                     NoValidConnectionsError)
@@ -302,7 +303,13 @@ def read_from_contentstore(course_key, path):
 def ssh_to(user, ip, key):
     ssh = paramiko.SSHClient()
     ssh.set_missing_host_key_policy(paramiko.AutoAddPolicy())
-    pkey = paramiko.RSAKey.from_private_key(StringIO(key))
+
+    pkey = None
+    for key_type in (RSAKey, Ed25519Key):
+        try:
+            pkey = key_type.from_private_key(StringIO(key))
+        except SSHException:
+            pass
 
     settings = get_xblock_settings()
     sleep_timeout = settings.get("sleep_timeout", 10)
--- a/hastexo/provider.py
+++ b/hastexo/provider.py
@@ -7,6 +7,7 @@ import paramiko
 import random
 import string
 import yaml
+from cryptography.hazmat.primitives import serialization, asymmetric
 
 try:
     from io import StringIO
@@ -15,7 +16,7 @@ except ImportError:
 
 from heatclient.exc import HTTPException, HTTPNotFound
 from keystoneauth1.exceptions.http import HttpError
-from novaclient.exceptions import ClientException
+from novaclient.exceptions import ClientException, NotFound
 from googleapiclient.errors import Error as GcloudApiError
 from googleapiclient.errors import HttpError as GcloudApiHttpError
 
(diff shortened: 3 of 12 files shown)


hastexo/provider.py
```diff
@@ -123,19 +124,36 @@ def set_environment(self, environment):
...
+            pkey = paramiko.RSAKey.generate(1024)
+            keypair["public_key"] = f'{pkey.get_name()} {pkey.get_base64()}'
+            s = StringIO()
+            pkey.write_private_key(s)
+            private_key = s.getvalue()
+            s.close()
 
         if encodeb64:
-            k = base64.b64encode(b(k))
+            private_key = base64.b64encode(b(private_key))
```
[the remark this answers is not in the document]
mrtmm: This encoding option is only used by the google cloud provider, I wanted to leave it as is just in case.  Should I change it still?

hastexo/provider.py
```diff
@@ -123,19 +124,36 @@ def set_environment(self, environment):
...
+                format=serialization.PublicFormat.OpenSSH).decode()
+            keypair["public_key"] = public_key
+
+            private_key = ed25519key.private_bytes(
+                encoding=serialization.Encoding.PEM,
+                format=serialization.PrivateFormat.OpenSSH,
+                encryption_algorithm=serialization.NoEncryption()).decode()
+
+        else:
+            pkey = paramiko.RSAKey.generate(1024)
```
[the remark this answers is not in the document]
mrtmm: Ack, done.

hastexo/provider.py
```diff
@@ -241,10 +259,22 @@ def get_stack(self, name):
         return {"status": status,
                 "outputs": outputs}
 
-    def create_stack(self, name, run):
+    def create_stack(self, name, run, key_type):
```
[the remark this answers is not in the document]
mrtmm: Yes, it should, thanks!

hastexo/common.py
```diff
@@ -302,7 +303,13 @@ def read_from_contentstore(course_key, path):
...
     ssh = paramiko.SSHClient()
     ssh.set_missing_host_key_policy(paramiko.AutoAddPolicy())
-    pkey = paramiko.RSAKey.from_private_key(StringIO(key))
+
+    pkey = None
+    for key_type in (RSAKey, Ed25519Key):
+        try:
+            pkey = key_type.from_private_key(StringIO(key))
+        except SSHException:
+            pass
```
fghaas, line 312: I'm not sure if this silent `pass` is a good idea. Isn't this a "can't happen" error? I mean, shouldn't the `key` argument, if passed in, _always_ be parseable as either an RSA or an ED25519 key? If that is the case, then I'd argue that you shouldn't catch this, and I'd suggest to remove the `try/except` block, and just throw any `SSHException` up the stack.
 
_If_ you want to allow for `key` being passed in as `None` or the empty string, then maybe do this instead?
```suggestion
    if key:
        for key_type in (RSAKey, Ed25519Key):
            pkey = key_type.from_private_key(StringIO(key))
```
Or am I missing something?

hastexo/provider.py
```diff
@@ -370,6 +404,11 @@ def delete_stack(self, name, wait=True):
         try:
             self.logger.info("Deleting OpenStack Heat stack [%s]" % name)
             self.heat_c.stacks.delete(stack_id=name)
+            self.nova_c.keypairs.delete(name)
+        except NotFound:
+            # Nova not finding a keypair to delete is a valid scenario,
+            # if the key is handled by heat via the lab template.
+            pass
```
fghaas, line 411: If we replace this `pass` with a call to `logger.info()`, then we get some traces in the logs if the key pair can't be found. This will come in handy when we eventually move from Heat-generated to XBlock-generated private keys, when such a log message reminds us of courses where the lab hasn't yet been updated.

fghaas (commented):
This looks great to me now. I just have two questions about exception handling. :)

hastexo/common.py
mrtmm: We don't know which type of key we have here, while we try to connect, so we should either try both and allow the "wrong" type to fail or we need to determine before which key we have and import accordingly.

fghaas: Meh, I'm an idiot. Complete thinko on my part here. 🫤

hastexo/provider.py
mrtmm: Sure, makes sense, done.

README.md
```diff
@@ -482,6 +482,11 @@ The following are optional:
 * `stack_template_path`: The static asset path to the orchestration template,
   if not specified per provider below.
 
+* `stack_key_type`: An SSH key type for accessing the lab environment.
+  Options are `rsa`, `ed25519` and `None`(default). If a key type is chosen,
```
fghaas, line 486: Micro-nit:
```suggestion
  Options are `rsa`, `ed25519` and `None` (default). If a key type is chosen,
```

hastexo/provider.py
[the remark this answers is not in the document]
fghaas: Nah, I guess you can leave it. :)

fghaas (commented):
I have one more microscopic nit, and then I think this is good to merge.

fghaas (approved):
Looks great, thanks!

fghaas:
@mrtmm If you have any last-minute second thoughts please do let me know; otherwise I would go ahead and merge this. :)

mrtmm:
I believe it's ready to merge, thanks!

PR #260, opened 2023-09-08
fix: Fix lab resume with an SSH key generated by the XBlock

mrtmm:
Make sure to fetch the key from the database when resuming a lab
stack.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,5 +1,8 @@
 Unreleased
 -------------------------
+* [Bug fix] Fix resuming a lab stack when the `stack_key_type`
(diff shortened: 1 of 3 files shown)


mrtmm:
@fghaas this is ready for review, although I am still trying to understand the test failures.

fghaas (commented):
This looks good to me! Re the test failures, see #261. If we merge that and then you rebase your topic branch on current `master`, I would expect the tests to pass.

fghaas (approved):
Tests pass now, so this looks good to merge as far as I am concerned. :)

mrtmm:
Ack, thanks!

PR #266, opened 2023-09-15
fix: Fix editing the stack_key_type in Studio

mrtmm:
Use an empty string instead on `None` as a default option for the `stack_key_type` attribute so Studio will allow resetting the value.

In addition, include the attribute in the Studio export.


fghaas (commented):
This still needs an update to `README.md`, please (which currently says the default for `stack_key_type` is `None`).

fghaas (approved):
Looks good to me now, thank you!

PR #268, opened 2023-11-17
feat: Add an option to launch a lab in a new window

mrtmm:
Add a new view to the XBlock for launching a lab in a new window. Consolidate the use of the guacamole-common-js library between the inline LMS lab and the lab launched in a new window.


hastexo/public/js/main.js
```diff
@@ -635,6 +446,9 @@ function HastexoXBlock(runtime, element, configuration) {
         terminal_client.disconnect()
 
         var dialog = $('#idle');
+        if (pause) {
+            dialog.find('.message').html('Your lab is currently active in a new window');
```
fghaas, line 450: Nit: I think calling it a "separate" window would be better. That's because the session may well be open for a period of, say, half and hour — at which point it's arguably not a "new" window anymore. :)

hastexo/static/html/main.html
```diff
@@ -9,6 +9,7 @@
     <select class="port"></select>
     <input type="button" value="Reset" class="reset" />
     <input type="button" value="Check progress" class="check" />
+    <input type="button" value="Open in new window" class="launch"/>
```
fghaas, line 12: Another nit: I'd love to keep the "Check Progress" button the rightmost. So I think we'll want the "Open in new window" button perhaps left-aligned below the terminal?

fghaas (commented):
Thanks! Looks like an excellent start to me; I just have two micro-nits. Other than the obvious human testing we'll still need to complete, the only thing I can think of that's actually missing is a README update and a changelog entry. :)

hastexo/public/js/main.js
```diff
@@ -307,6 +173,25 @@ function HastexoXBlock(runtime, element, configuration) {
             dialog.dialog(dialog_container);
         };
 
+        terminal_client.onidle = function(pause) {
+            /* We're idle; clear stack info.  */
+            stack = null;
+    
+            var dialog = $('#idle');
+            if (pause) {
+                dialog.find('.message').html('Your lab is currently active in a separate window');
```
fghaas, line 182: This could still use a UX improvement. :) Currently, this ends up saying:

> We think you're busy elsewhere.
> 
> Your lab is currently active in a separate window
>
> Click 'OK' to resume it.

Apart from the fact that we're missing a full-stop after "window", it's also not quite the right message. By saying "Click 'OK' to resume it." we suggest that the lab is suspended, which it clearly isn't.

Maybe we can fix this by saying "Click 'OK' to return to your terminal in this window.", which would fit the pop-out use-case and would *also* apply to resuming from suspend?

fghaas (commented):
This is working very well as far as I can tell; I have a minor suggestion about what we're displaying to the user.

hastexo/static/html/main.html
mrtmm: Done.

hastexo/public/js/main.js
mrtmm: Sure, done.

fghaas (approved):

PR #270, opened 2023-11-30
fix: Fix private key getting lost on stack resume failure

mrtmm:
Starting with v7.7.1, we are fetching the private key from the database. In case of a resume failure, we need to keep the key in place when running the cleanup step after the failed resume so that the learner will not lose access to the lab altogether.

### DIFF
--- a/hastexo/tasks.py
+++ b/hastexo/tasks.py
@@ -255,12 +255,6 @@ class LaunchStackTask(HastexoTask):
         except LaunchStackFailed as e:
             logger.error(e.error_msg)
 
-            # In case of failure, only return the provider if this was a failed
-            # resume attempt.
-            provider_name = ""
-            if e.suspend:
-                provider_name = e.provider.name
-
             stack_data = {
                 'status': e.status,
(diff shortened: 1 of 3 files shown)


hastexo/tasks.py
```diff
@@ -266,7 +266,7 @@ def run(self, **kwargs):
                 'error_msg': textwrap.shorten(e.error_msg, width=256),
                 'ip': None,
                 'user': "",
-                'key': "",
+                'key': self.stack_key if e.suspend else '',
```
fghaas, line 269: This looks good to me, but I think Future Maari (and anyone else reviewing this bit of code in the future) will appreciate a brief explanation _why_ this logic applies here. :) Perhaps add a comment to that effect right above the `stack_data` definition in line 264?

mrtmm: Fair point. And on second thought, I was thinking that perhaps it's better to not touch the key at all, in case of a failed resume. So instead of overwriting it with the same key that's already there, remove it from the fields to be updated.

PR #272, opened 2023-12-13
Bump version: 7.7.3 → 7.8.0

mrtmm:


fghaas (approved):
I was debating for a bit if we should go with the major version bump right away, but it's probably smarter to leave that for the version whose dependencies are aligned with Open edX Quince. So, this looks good to merge to me. :)

PR #280, opened 2024-02-13
Check for unauthorized access for fullscreen labs

mrtmm:
Don't try to access a fullscreen lab when a user is unauthenticated. This would send out a query with empty parameters and fail anyway; Instead, check if a user is authenticated right away and return a 401 if not.


hastexo/hastexo.py
```diff
@@ -658,6 +658,10 @@ def launch_new_window(self, request, suffix=''):
         """
         The fullscreen lab view, opened in a new browser window.
         """
+
+        if not request.cookies.get('edxloggedin', False):
```
fghaas, line 662: Just checking: are you sure you want to test for the existence of one specific cookie here, rather than generically checking whether the request is associated with a valid session? Does `request.session` give you a generic way to do that? (I'm not sure if it does.)

mrtmm: No, there is no session option, it's a [webob.Request](https://docs.pylonsproject.org/projects/webob/en/stable/api/request.html#module-webob.request). We can check for `edx-user-info`, `edxloggedin` or `sessionid` cookies. I can switch to checking if `sessionid` is present.

fghaas: Maybe go for `sessionid`, then. Who knows if `edxloggedin` doesn't become `openedxloggedin` at some point. :)

mrtmm: Ack, done.

fghaas (commented):
This looks good to me now, thank you!

fghaas (approved):

PR #282, opened 2024-03-28
Add Python 3.12 support

mrtmm:
-   Update test matrix:
    -  xblock-sdk v0.9.0 includes a breaking change for XBlock versions before 2.0 but there is no upper constraint included in the requirements for this for all XBlock versions. Thus, all our tests have been failing since the release of xblock-sdk==0.9.0 on 2024-03-21.
    -   Update our test matrix to include XBlock versions 1.7 / 1.8 / 1.9 and add an upper constraint for xblock-sdk.
    -   The Open edX Quince release is using XBlock v1.8, which does pin the xblock-sdk to v0.7.0 so in this case we only need to add the constrait to our test runs.

- Add support for Python 3.12

### DIFF
--- a/.github/workflows/tox.yml
+++ b/.github/workflows/tox.yml
@@ -17,6 +17,10 @@ jobs:
         pip-version:
           - 22.0.4
           - 23.0.1
+          - 23.2.1
+        include:
+          - python-version: '3.12'
+            pip-version: '23.2.1'
 
     steps:
     - name: Check out code
@@ -27,7 +31,7 @@ jobs:
         python-version: ${{ matrix.python-version }}
     - name: Install dependencies
       run: |
-        pip install "tox<4" tox-gh-actions tox-pip-version
+        pip install "tox<4" tox-gh-actions tox-pip-version setuptools
     - env:
         TOX_PIP_VERSION: ${{ matrix.pip-version }}
       name: Test with tox (pip ${{ matrix.pip-version }})
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,3 +1,7 @@
+Unreleased
+-------------------------
+* [Enhancement] Add support for Python 3.12.
+
 Version 7.9.1 (2024-02-14)
 -------------------------
 * [Bug fix] Don't allow accessing a fullscreen lab directly from a URL,
--- a/hastexo/__init__.py
+++ b/hastexo/__init__.py
@@ -1,8 +1,8 @@
-import pkg_resources
+from importlib import metadata
 # __version__ attribute as suggested by (deferred) PEP 396:
 # https://www.python.org/dev/peps/pep-0396/
 #
 # Single-source package definition as suggested (among several
 # options) by:
 # https://packaging.python.org/guides/single-sourcing-package-version/
-__version__ = pkg_resources.get_distribution('hastexo-xblock').version
+__version__ = metadata.version('hastexo-xblock')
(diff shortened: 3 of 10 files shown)


fghaas:
I believe that to unbreak the tests on Python 3.12, you'll still need to add `setuptools` to the `requirements/tests.txt` list. But maybe you can achieve the same effect by dropping the `<6` version specifier on `setuptools-scm` in `requirements/setup.txt` — the reason why we had that constraint in the requirements (https://github.com/mrtmm/hastexo-xblock/commit/e2dc395e931af343d5507a5ff9205d0d147fd928) is long obsolete.

requirements/test.txt
```diff
@@ -18,3 +18,4 @@ path.py>=12.4
 ddt
 nose
 coverage
+setuptools-scm
```
fghaas, line 21: This addition should not be necessary, now that we've added `setuptools-scm` to `base.txt` and we're already saying `-r base.txt` at the top of this file.

.github/workflows/tox.yml
```diff
@@ -27,7 +29,7 @@ jobs:
         python-version: ${{ matrix.python-version }}
     - name: Install dependencies
       run: |
-        pip install "tox<4" tox-gh-actions tox-pip-version
+        pip install "tox<4" tox-gh-actions tox-pip-version setuptools-scm
```
fghaas, line 32: I *think* this addition is also not necessary, since we do pull in `setuptools-scm` via `setup.py` and `base.txt` now.

.github/workflows/tox.yml
```diff
@@ -14,9 +14,11 @@ jobs:
           - '3.9'
           - '3.10'
           - '3.11'
+          - '3.12'
         pip-version:
-          - 22.0.4
-          - 23.0.1
+          # - 22.0.4
+          # - 23.0.1
+          - 23.2.1
```
fghaas, line 21: This makes me think that we should probably replace this matrix with something else. It doesn't make a ton of sense to run the latest Python with an age-old `pip`, just like it doesn't make sense to run an ancient Python with bleeding-edge `pip`.

fghaas (commented):
Just a few thoughts here.

.github/workflows/tox.yml
fghaas: A proposed fix for this is in https://github.com/mrtmm/hastexo-xblock/pull/8.

requirements/test.txt
mrtmm: we only have `setuptools-scm` in `setup.txt` at the moment.

.github/workflows/tox.yml
mrtmm: Ack, thanks, I pulled that in and tests are passing :)

requirements/test.txt
mrtmm: I moved the `setuptools-scm` to `base.txt` now and removed from here.

.github/workflows/tox.yml
mrtmm: I tried again, and even with `setuptools-scm` in `base.txt`, gh-actions was missing this and all tests failed. So, I think we need this addition here.

fghaas: I completely understand that I'm nitpicking here, but I *think* https://github.com/mrtmm/hastexo-xblock/pull/9 has a more correct fix for this. :) What do you think?

mrtmm: Sure, looks good. I merged and squashed it, not sure why the PR didn't close, I think I should have pushed the branch before squashing it.

fghaas (approved):
Looks good to me now, thanks!

PR #283, opened 2024-04-12
Improve error handling for missing Provider configurations

mrtmm:
Throw a ProviderException when a provider configuration is missing but the Provider is linked to an existing Stack.

### DIFF
--- a/hastexo/provider.py
+++ b/hastexo/provider.py
@@ -66,6 +66,9 @@ class Provider(object):
                 return OpenstackProvider(name, config, sleep_seconds)
             elif provider_type == "gcloud":
                 return GcloudProvider(name, config, sleep_seconds)
+        else:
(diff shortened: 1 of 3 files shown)


hastexo/provider.py
```diff
@@ -66,6 +66,9 @@ def init(name):
                 return OpenstackProvider(name, config, sleep_seconds)
             elif provider_type == "gcloud":
                 return GcloudProvider(name, config, sleep_seconds)
+        else:
```
fghaas, line 69: I think it's suboptimal that we end up with the same exception type with the same message in two different code paths, as it complicates troubleshooting. Thoughts?

mrtmm: Hmm, perhaps yes. I updated the error message now.

PR #285, opened 2024-04-22
Stop installing the XBlock in editable mode

mrtmm:
Drop the `-e .` from `requirements/base.txt`


fghaas (approved):
Looks good to me, thanks!

PR #289, opened 2024-07-23
Add support for the Open edX Redwood release

mrtmm:


Changelog.md
```diff
@@ -1,3 +1,13 @@
+Unreleased
+-------------------------
+* [Enhancement] Update requirements to keep in sync with `edx-platfom` 
+  and support the Open edX Redwood release.
+* [BREAKING CHANGE] Update the import statements for `xblock-utils`.
```
fghaas, line 5: Just checking: can't we make this a non-breaking change by using a `try`/`except` block on the import statement and catching the `ImportError`?

fghaas:
We have some tests breaking here (complaining about `StudioContainerWithNestedXBlocksMixin`) — is this an XBlock dependency version that still needs fixing, as in https://github.com/citynetwork/markdown-xblock/pull/44?

mrtmm:
Yes, I am still working on this, I reverted to `Draft` until it's ready to be reviewed.

fghaas (approved):

fghaas:
Sorry, hold on, I fat-fingered my approval and I can't find a quick way to undo it.

requirements/base.txt
```diff
@@ -17,10 +17,10 @@ python-novaclient>=7.1.2,<16
 tenacity>=6.2,<8
 
 # for hastexo_guacamole_client
-django<=4.2.8
+django<=4.2.14
 channels<=4.0.0
 daphne<=4.0.0
 twisted<24 # drop this restriction once we drop Python 3.8 and 3.9 support
```
fghaas, line 23: Should we perhaps use a `python_version` environment marker here too, as we do for `xblock-sdk` in `test.txt`?

fghaas (approved):

requirements/base.txt
```diff
@@ -17,10 +17,10 @@ python-novaclient>=7.1.2,<16
 tenacity>=6.2,<8
 
 # for hastexo_guacamole_client
-django<=4.2.8
+django<=4.2.14
 channels<=4.0.0
 daphne<=4.0.0
-twisted<24 # drop this restriction once we drop Python 3.8 and 3.9 support
-mysqlclient<=2.2.1  # keep in sync with edx-platform
+twisted<24;python_version<="3.9" # drop this restriction once we drop Python 3.8 and 3.9 support
```
fghaas, line 23: Sorry, no, it doesn't work this way. This means "install `twisted`, using a version before 24.0.0, but only on Python releases earlier than 3.9, *and don't install `twisted` at all* for later releases." That's not what we want. Instead, we want:
```suggestion
twisted<24;python_version<"3.10"
twisted;python_version>="3.10"
```

mrtmm: But don't we want for Python versions later than 3.9 to install it via Daphne/Channels as we did before? https://github.com/hastexo/hastexo-xblock/commit/eccd8a4b8b1fb8752f198508cb38e307c204e684

fghaas: Ah, sorry, so we do get `twisted`, just via `daphne` and not directly. OK, sounds good to me then. Thanks for the reminder. :)

fghaas (approved):

PR #292, opened 2024-09-25
Bump GitHub Actions version references

fghaas:
Bump the tag reference for the following GitHub Actions to their latest versions:

- `actions/upload-artifact`
- `actions/download-artifact`
- `actions/checkout`
- `actions/setup-python`
- `codecov/codecov-action`

This fixes the `report` job.

### DIFF
--- a/.github/workflows/tox.yml
+++ b/.github/workflows/tox.yml
@@ -35,11 +35,12 @@ jobs:
       name: Test with tox (pip ${{ matrix.pip-version }})
       run: tox
     - name: Upload coverage data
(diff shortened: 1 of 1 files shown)


fghaas:
@mrtmm Right now there's nothing I can do about the `codecov/project` check failure, since several changes landed in `master` without that specific check, so we now have a reduction in coverage against *what Codecov has recorded* for the current state of `master`. I thus intend to override that check, when you've added your approval.

mrtmm (approved):
Ack, thanks!

PR #297, opened 2024-11-13
Restrict access to fullscreen labs directly from an URL

mrtmm:
Add more authentication checks when acessing a fullscreen lab direclty from an URL. Return a 401 when the user is not authenticated in the platfrom.

### DIFF
--- a/hastexo/hastexo.py
+++ b/hastexo/hastexo.py
@@ -663,7 +663,11 @@ class HastexoXBlock(XBlock,
         """
         The fullscreen lab view, opened in a new browser window.
         """
-        if 'sessionid' not in request.cookies:
+
+        # Properly authenticated requests contain two session cookies:
+        # sessionid and edxloggedin.
+        # If either of them isn't present, return HTTP 401 immediately.
+        if not all(c in request.cookies for c in ('sessionid', 'edxloggedin')):
             return Response(status=401, body="Unauthorized")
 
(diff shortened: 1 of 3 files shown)


hastexo/hastexo.py
```diff
@@ -663,7 +663,7 @@ def launch_new_window(self, request, suffix=''):
         """
         The fullscreen lab view, opened in a new browser window.
         """
-        if 'sessionid' not in request.cookies:
+        if not all(c in request.cookies for c in ('sessionid', 'edxloggedin')):
```
fghaas, line 666: I had a bit of trouble parsing what this does, so I'd suggest to add a comment to benefit our future selves. :)

```suggestion
        # Properly authenticated requests contain two session cookies: sessionid and edxloggedin.
        # If either of them isn't present, return HTTP 401 immediately.
        if not all(c in request.cookies for c in ('sessionid', 'edxloggedin')):
```

fghaas (commented):
Looks great. I have just one suggestion.

hastexo/hastexo.py
mrtmm: Ack, done.

fghaas: Thank you!

fghaas (approved):
Looks great to me now! Please go ahead and merge.

PR #304, opened 2025-02-05
Add support for resizing the lab window

mrtmm:
Add an observer to the JavaScript code to detect lab size changes on the LMS page and send the new dimensions to the guacamole client.

### DIFF
--- a/README.md
+++ b/README.md
@@ -198,6 +198,15 @@ This is a brief explanation of each:
 * `enable_fullscreen`: An option to allow learners to launch a lab in fullsceen mode,
   in a separate browser window. (Default: `false`)
 
+* `lab_resize_method`: A [display setting](https://guacamole.apache.org/doc/1.5.5/gug/configuring-guacamole.html#rdp-display-settings) for RDP connections to update the RDP server
+  when the width and height of the client display changes. Possible values are:
+
+  * `display-update` *Only supported with Windows RDP 8.1 or xrdp 0.10.1 (and higher)*. 
+    Uses the "Display Update" channel to request the server to change the display size.
+
+  * `reconnect` (the default) Automatically disconnects the RDP session when the client 
+    display size has changed, and reconnects with the new size.
+
 * `lab_usage_limit`: Allocate limited time per user for labs across the platform,
   in seconds. (Default is `None`, meaning there is no time limit).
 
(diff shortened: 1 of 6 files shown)


hastexo_guacamole_client/consumers.py
```diff
@@ -33,20 +33,24 @@ async def connect(self):
...
-            hostname=stack.ip,
-            port=params.get('port', [default_port])[0],
-            username=stack.user,
-            password=stack.password,
-            private_key=stack.key,
-            color_scheme=settings.get("terminal_color_scheme"),
-            font_name=settings.get("terminal_font_name"),
-            font_size=settings.get("terminal_font_size"),
-        )
+        self.client.handshake(connection_args)
```
fghaas, line 53: I may be making a fool of myself here because something might have changed in Python ages ago, and I may have missed it, but looking at [`GuacamoleClient.handshake()`](https://github.com/mohabusama/pyguacamole/blob/master/guacamole/client.py#L138) shouldn't this line look like so?

```suggestion
        self.client.handshake(**connection_args)
```

mrtmm: Yes, that was a typo. But, I believe we don't need to split the arguments between different protocols at all, SSH does not have the "resize-method" argument and will simply ignore it.

README.md
```diff
@@ -198,6 +198,15 @@ This is a brief explanation of each:
 * `enable_fullscreen`: An option to allow learners to launch a lab in fullsceen mode,
   in a separate browser window. (Default: `false`)
 
+* `lab_resize_method`: A [display setting](https://guacamole.apache.org/doc/1.5.5/gug/configuring-guacamole.html#rdp-display-settings) for RDP connections to update the RDP server
+  when the width and height of the client display changes. Possible values are:
+
+  * `display-update` *Only supported with RDP 8.1 and higher*. Uses the "Display Update"
+    channel to request the server to change the display size.
```
fghaas, line 205: "RDP 8.1" refers to the Windows RDP implementation, which does not map directly to supported features in xrdp. As far as I have been able to make out from https://github.com/neutrinolabs/xrdp/issues/1272, https://github.com/neutrinolabs/xrdp/pull/2891, and https://github.com/neutrinolabs/xrdp/blob/v0.10.2/NEWS.md, GFX support was implemented in xrdp v0.10.0 (2024-05-10), and a resize bug was subsequently fixed in xrdp v0.10.1 (2024-07-31).

What do you think? Should we perhaps change this to:

```suggestion
  * `display-update` *Only supported with Windows RDP 8.1 or xrdp 0.10.1 (and higher)*.
     Uses the "Display Update" channel to request the server to change the display size.
```

mrtmm: Yes, sounds good, thanks!

fghaas (approved):
Looks good to me, thank you!

PR #306, opened 2025-05-21
Bump upper bounds for dependencies, include package data

mrtmm:
* Update the install_requires list in setup.py to make sure the correct versions of requirements are installed with the XBlock.
* Add `include_package_data=True` in setup.py to make sure all package files will be installed.
* Include upper constraint for oslo.config and paramiko for both, setup.py install_requires list and requirements.txt.


setup.py
```diff
@@ -57,20 +57,30 @@ def package_data(pkg, roots):
...
+        'dogpile.cache<=1.3',
+        'google-api-python-client<=2.149.0',
+        'keystoneauth1<=5.8',
+        'openstacksdk<=4.1',
+        'osc-lib<=3.1',
+        'os-client-config<=2.1',
+        'oslo.serialization<=5.5',
+        'oslo.utils<=7.4',
+        'oslo.config<=9.7.1',
+        'paramiko>=3.4.0',
```
fghaas, line 71: Just checking: is there a specific reason why we're defining an upper bound for everything else, but only a lower bound for paramiko?

fghaas (commented):
Looks good in general, just one minor question. Also, I'd suggest to change the commit message subject to something like "Bump upper bounds for dependencies, include package data".

setup.py
mrtmm: The lower bound we set for vulnerability fix in https://github.com/hastexo/hastexo-xblock/pull/287/commits/9d477f909c3dcf3edeccb4b3f4637d407a893b7e
I am not sure why we don't have an upper bound in this case.

fghaas: Okay, in that case let's add `<4`.

mrtmm: Ack, done.

fghaas (approved):
Excellent, thank you!

PR #309, opened 2025-05-21
docs: Update README for Sumac

fghaas:
Mention Open edX Sumac in the README, and recommend using version 8.4
and up of the XBlock.


mrtmm (approved):
I missed the notification, thanks for the reminder!

PR #311, opened 2025-06-18
feat: Improve coverage for VNC

fghaas:
* Add a `wait_for_vnc()` method (analogous to `wait_for_rdp()`).
* Adapt the Guacamole client so that it uses 5900 as the default VNC port.
* Add VNC coverage to tests.

### DIFF
--- a/hastexo/tasks.py
+++ b/hastexo/tasks.py
@@ -570,11 +570,7 @@ class LaunchStackTask(HastexoTask):
 
         return ssh
 
-    def wait_for_rdp(self, stack_ip):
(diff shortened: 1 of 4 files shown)


fghaas:
@mrtmm This is extremely tentative, but does the approach here look reasonable to you?

(I'm debating if this is properly `feat` or `fix` — our README does mention VNC so there's some merit in the argument that this is really a fix.)

mrtmm:
Changes look good to me; I would keep the `feat`, a `fix` in my opinion would be updating the README to document the _actual_ feature support.

PR #315, opened 2025-09-18
Add upper constraint to oslo.i18n dependency

mrtmm:
Add upper constraint to oslo.i18n dependency to avoid pulling in a newer, incompatible version.


fghaas (approved):
Looks good to me, thank you!

PR #335, opened 2026-05-13
Add Mac Cmd+V paste support for the lab terminal

mrtmm:
Previously, using the Cmd+V key combination would create the paste event but also type a literal 'v' in the terminal. The keystroke is now suppressed while keeping the paste event, consistent with the existing Ctrl+V, Ctrl+Shift+V behaviour.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,6 +1,7 @@
 Unreleased
 -------------------------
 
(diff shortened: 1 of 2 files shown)


fghaas (dismissed):

fghaas:
Since you're already updating the PR, would you mind fixing the typo in your commit message subject? :)

fghaas (approved):

PR #344, opened 2026-06-15
chore: Drop Google-related dependencies

fghaas:
We dropped support for the GCloud provider in a3d5371df22140571b3f7bfd213a00540c1e9cc1.

While other parts of Open edX might still require Google-related libraries (and they may thus be installed in an Open edX image alongside the XBlock), there is really no longer any need to track these dependencies from this XBlock.

Thus, drop them from `pyproject.toml` and `requirements.txt`.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -9,7 +9,6 @@ dependencies = [
     "apscheduler<4",
     "cliff<=4.7",
     "dogpile.cache<=1.3",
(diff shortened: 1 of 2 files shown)


mrtmm (approved):

fghaas:
Superseded by #350.

fghaas:
Fixed in #350, closing.

PR #346, opened 2026-06-18
Fix translations for supported languages

mrtmm:
Move translations to the `/locale` directory as
`<lang_code>/LC_MESSAGES/django.po` and
`<lang_code>/LC_MESSAGES/djangojs.po`

Compile the `*.mo` files during the `openedx` image build via the
`tutor-contrib-stackamole` plugin instead.

Update HACKING.md


fghaas (approved):
Looks good to me, thanks!

